PR #73, opened 2014-04-12
rack-mini-profiler should use Rails URL helpers when running under rails

### LINKED ISSUE #25: Does not work on pages handled by a mounted engine
For pages handled by a mounted engine, the inserted script element looks like this

    <script async="" type="text/javascript" id="mini-profiler" src="/foo-engine/mini-profiler-resources/includes.js?v=..." ...></script>

Note the source attribute, it is prefixed by "/foo-engine" (the mythical example engine). Unfortunately, rmp does not provide its script at that locating. The script is supposed to be at /mini-profiler-resources/includes.js.

As far as I can tell, the reason for the wrong path is that Rails sets env['SCRIPT_NAME'] to point at the engine's mount point. That parameter in turn is used by rmp to set up its script path.

A very hackish woraround I'm currently using is this change in Rack::MiniProfiler#call

      #return serve_html(env) if path.start_with? @config.base_url_path
      return serve_html(env) if path.include? @config.base_url_path

cbeer:
When running under Rails, rack-mini-profiler should use the Rails
`#url_for` helper in order to correctly construct the path to the
profiler javascript when in a Rails engine context where SCRIPT_NAME is
inappropriate.

Fixes #25

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -588,7 +588,11 @@ Append the following to your query string:
     # * you have disabled auto append behaviour throught :auto_inject => false flag
     # * you do not want script to be automatically appended for the current page. You can also call cancel_auto_inject
     def get_profile_script(env)
-      path     = "#{env['RACK_MINI_PROFILER_ORIGINAL_SCRIPT_NAME']}#{@config.base_url_path}"
+      path = if env["action_controller.instance"]
+        env["action_controller.instance"].url_for("#{@config.base_url_path}")
+      else
+        "#{env['RACK_MINI_PROFILER_ORIGINAL_SCRIPT_NAME']}#{@config.base_url_path}"
+      end
 
       settings = {
        :path            => path,
--- a/spec/integration/mini_profiler_spec.rb
+++ b/spec/integration/mini_profiler_spec.rb
@@ -134,6 +134,20 @@ describe Rack::MiniProfiler do
 
   end
 
+  describe 'within a Rails application' do
+
+    before do
+      mock_controller = double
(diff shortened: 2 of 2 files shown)


prakashmurthy:
I ran into the same issue with a mounted rails engine, and checked out both the solutions suggested in this thread. 

Both solutions work for me in both the main app and the mounted engine. 

The change suggested by @superscott makes the https://github.com/MiniProfiler/rack-mini-profiler/blob/7a56508b65480d2e0ce322775e71a9d78e21f483/spec/integration/mini_profiler_spec.rb#L132-L140 test fail.

luizdamim:
I just started a new project using Rails Engines and I'm having this same issue. How can I help to get this patch merged? :)

SamSaffron:
@luizdamim we need to get this rebased for a start and we got to confirm it does not lead to any breakage.

superscott:
I don't have a handy environment to test Rails Engines currently, but it looks like @cbeer 's changes https://github.com/cbeer/rack-mini-profiler/commit/a8839d06fcbf18baf9ba678e4db6f0f9e688e55f should fix it.

ericboehs:
:+1:

Brit200313:
I just wanted to chime in that I have a Rails Engine loaded into a Rails Engine loaded into an app. This PR worked perfectly for me. Thank you!

SamSaffron:
k   I will merge it in @cbeer can you rebase?

cbeer:
@SamSaffron rebased. I'm not actively using this gem, so it would be good if someone could just double-check the work? @Brit200313, could you maybe take another look?

Brit200313:
@cbeer @SamSaffron still works for me. Cheers!

SamSaffron:
K lets try this! thanks

PR #85, opened 2014-06-13
Angularjs compat

nathantsoi:
AngularJS compatibility

since is sometimes emptied before mini profiler can be initialized, we need to put the script tag in the head.

set ```Rack::MiniProfiler.config.inject_into = 'head'``` in config/initializers/mini_profiler.rb in your angular rails app


nathantsoi:
Should be all set.

SamSaffron:
sorry one last thing here, 

inject into should be a symbol eg `:head` or `:body` feels really odd as a string. can you change that?

nathantsoi:
good call

shanbhardwaj:
@nathantsoi @SamSaffron guys it would be great to have this working.

nathantsoi:
@shanbhardwaj, I've fixed this by moving the ng-view declaration inside the body

joshuaflanagan:
Sounds like this PR should be closed, and replaced with a caveat in the README directing people to use the workaround (stop making `body` your `ng-view`). Is that right?

nathantsoi:
yep!

PR #91, opened 2014-07-10
fix for backwards compatibility with ruby 186

wavesummit:
cannot use define_method with a block


kbrock:
Is this still a valid request?

SamSaffron:
Sorry, closing this, ruby 1.8.7 is not something I support supporting anymore.

PR #103, opened 2014-10-21
Add web service metrics with Savon integration

nicedawg:
Admittedly this is a quick and dirty pull request and probably shouldn't be merged in unchanged. 

There are things that should be refactored, and this doesn't consider the desire to reorganize the internal classes to match the UI spec. 

However, if you want Savon integration so that Web Service calls are included in the MiniProfiler output, this will get you there in the meantime! Let me know if you see any way this should be improved.


SamSaffron:
This is conflicting and way too giant for me to include, we need to figure out a way for you to add this in without impacting the core of mini profiler so much.

PR #132, opened 2015-03-23
UI Submodule

thomas-mcdonald:
Just throwing an *early* PR up for my work on switching the mini profiler to use the UI repository

Stuff to do:
- [x] fix tests
- [ ] refactor struct classes?
- [x] finish converting all request data into UI format - fairly sure some of this can be trimmed off (looks like time without child duration is now computed clientside, for example)
- [ ] switch to using `include.partial.html` over `profile_handler.js`


SamSaffron:
wow, awesome, excited about this work

thomas-mcdonald:
@SamSaffron pretty roughly 'done'. Would probably hang off refactoring structs too much until this is merged. things I'm not sure about/uncomfortable with:

* `:has_user_viewed` key in the Page struct, looks like it should maybe be a property since it shouldn't be  returned to the client? Similarly with the `:user` key
* I pulled the `type` attribute off of custom structs, since it seems to have no value except for setting the column name in `CustomTimings`
* Strange capitalised keys everywhere :frowning:

Aside from that, the last thing that needs doing it switching the html file/options that are inserted for the `<script>` tag, but I've done far too much procrastinating for one evening.

SamSaffron:
Can you make sure no new functionality that we added since the fork is
lost?

For example recently I added the ability to highlight slow queries in the
query list (which needs to be ported back into the UI project) I can work
with you on merging all the fixes / changes into UI first before we take on
the dependency.

thomas-mcdonald:
Will take a look at it over the next few days

SamSaffron:
hmm any update here?

thomas-mcdonald:
been busy, hoping to push this into one of our apps in the next week or so - may give me a chance to revisit

SamSaffron:
wow, awesome, I am looking forward to this refresh

kbrock:
This looks like there are a few cherry-pickable commit in here.

Is there still desire to:

1. Change hash keys to camel case?
2. Move ui to another repo using a submodule?
3. extract sql timings?

Point to a commit number you want, and I can pull out into a separate PR (keeping attribution of course)

SamSaffron:
Initially I was very eager to just have one UI module. but as time has
passed I am not 100% sure this is a correct course of action, Ruby
profiling has its own special sauce and its going to be a nightmare
managing change here.

I think a better approach here is probably to move us to a more modern
templating language like say handlebars and have it precompiled. also we
should review feature missing in ui submodule and look at pulling them in.

Issue #114, opened 2014-12-05
Mounted Routes


kamilpogo:
rack-mini-profiler doesn't work with mounted routes like
```ruby
namespace :admin do
  mount Resque::Server, :at => 'resque'
end
```
It inject html that tries to load "/admin/resque/mini-profiler-resources/includes.js" which cannot be found because it serves "/mini-profiler-resources/includes.js"

albertodega:
same here! +1

TrevorS:
:+1:

brodock:
:+1: same here

ballou88:
This seems to happen for engines that have mounted routes.  It was introduced with https://github.com/MiniProfiler/rack-mini-profiler/commit/29aac5d4.

I don't fully understand what this commit was trying to accomplish.  If anyone can give me a little insight maybe I can work up a pull request.  Thanks.

smarterclayton:
It was to support rack apps that were not running at the root, ie within passenger you can define a prefix string (/app) that when your rack app gets invoked (/app/bar), the request path rack sees is /bar and an env SCRIPT_NAME "/app".  This is supported in Rails, but somewhat spotty.  I don't have an immediate guess why the namespace would break it - I believe at the time we also were using namespaces but may not have been a mounted engine under a namespace.

brodock:
any updates on the issue?
It looks like that even skip_paths are not working for this specific issue.

haggen:
Almost a year and nothing? :'(

SamSaffron:
I am sad too, but someone is going to need to work on this I merged one attempted fix in this department and it is already showing cracks. 

Subdirectory installs in rails and rack follow weird conventions

PR #133, opened 2015-03-30
Save and use original SCRIPT_NAME before calling an @app. Fix #25 and #114 + test

### LINKED ISSUE #25: Does not work on pages handled by a mounted engine

907th:
The issue is that Rails engines break `env['SCRIPT_NAME']`. So we just need to store it somewhere before calling `@app.call`.

Fixed #25 and #114.

CC  @mschuerig @kamilpogo @smarterclayton

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -100,7 +100,7 @@ module Rack
 
         # Otherwise give the HTML back
         html = MiniProfiler.share_template.dup
-        html.gsub!(/\{path\}/, "#{env['SCRIPT_NAME']}#{@config.base_url_path}")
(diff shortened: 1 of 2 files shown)


SamSaffron:
this is one odd bug :) thanks for sorting it out!

PR #137, opened 2015-04-29
Use jQuery in MiniProfiler namespace

helsus:
New version of https://github.com/MiniProfiler/rack-mini-profiler/pull/135

 `jQuery.noConflict` with true argument cleans up both `$` and `jQuery`:
https://api.jquery.com/jquery.noconflict/#jQuery-noConflict-removeAll
https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/html/includes.js#L703
```javascript
 load(options.path + "jquery.1.7.1.js?v=" + options.version, function() {
    MiniProfiler.jQuery = $ = jQuery.noConflict(true);
    $(deferInit);
});
```

So, if page does't have already loaded jQuery there will be an error:
https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/html/includes.js#L666
```javascript
if (jQuery.hotkeys === undefined) {
```


SamSaffron:
nice, yeah, much better

PR #140, opened 2015-05-21
Update misleading comment in railtie.rb

rthbound:
There is no #post_authorize_cb

The comment here should likely be amended (or removed entirely if the statement is no longer true).


SamSaffron:
the comment should point to `Rack::MiniProfiler.authorize_request` , if called in prd it enables profiling.

rthbound:
@SamSaffron I see... Something more like:

rthbound:
@SamSaffron I see... I've updated the PR to amend the comment rather than removing it entirely.

How's this?

SamSaffron:
looks good to me! thanks

PR #141, opened 2015-06-07
Check for dependencies with defined? rather than NameError

dgynn:
This is consistent with the checks for Net::HTTP and ::Rails and reduces memory allocated by ActiveSupport handling the exceptions.


dgynn:
That works for me. I'll submit a new pull request with it done that way. Thanks.

PR #142, opened 2015-06-15
Check for dependencies with defined? before checking class_exists?

dgynn:
This is a re-do of #141 that keeps the original class_exists? logic but adds the check for defined?(Class) first to prevents NameErrors during loading.


dgynn:
Thanks @SamSaffron and thanks for MemoryProfiler which showed me where that memory allocation was happening on startup.

SamSaffron:
awesome!

PR #151, opened 2015-07-30
Truncate strings in the analyze-memory report

nateberkopec:


nateberkopec:
Not sure what to do on this line: https://github.com/nateberkopec/rack-mini-profiler/blob/truncate-analyze-memory/lib/mini_profiler/profiler.rb#L524

What's `trunc` for?

SamSaffron:
looks good, thanks

PR #154, opened 2015-08-09
Make authorized admin example code actually work

schnittchen:
Also, don't introduce a so generally named method like "authorize" into every controller. People blindly copy&paste code from READMEs.


SamSaffron:
thanks

PR #156, opened 2015-08-16
Add mongo/mongoid5 support

nofxx:
Mongoid 5 dropped moped and it's using default mongo driver.
This adds  support for it. 

Note:
I'm using ActiveSupport's method_chain because Mongoid already requires it, may change if you want.

### DIFF
--- a/lib/patches/db/mongo.rb
+++ b/lib/patches/db/mongo.rb
@@ -1,0 +1,12 @@
+# Mongo/Mongoid 5 patches
+class Mongo::Server::Connection
+  def dispatch_with_timing(*args, &blk)
+    return dispatch_without_timing(*args, &blk) unless SqlPatches.should_measure?
+
+    result, _record = SqlPatches.record_sql(args[0][0].payload.inspect) do
+      dispatch_without_timing(*args, &blk)
+    end
+    return result
+  end
+  alias_method_chain :dispatch, :timing
+end
--- a/lib/patches/sql_patches.rb
+++ b/lib/patches/sql_patches.rb
@@ -39,6 +39,7 @@ end
 
 require 'patches/db/mysql2'         if defined?(Mysql2::Client) && SqlPatches.class_exists?("Mysql2::Client")
 require 'patches/db/pg'             if defined?(PG::Result) && SqlPatches.class_exists?("PG::Result")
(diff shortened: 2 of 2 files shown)


SamSaffron:
sure, let try this out, thanks!

nofxx:
@SamSaffron I missed something... =/

I was using all day yesterday `#inspect` which give all the info but it's a lil messy:

```
#<Mongo::Protocol::Query:0x0000000727c200 @database="database", @namespace="database.collections", @selector={"_id"=>"xxxxxx"}, @options={:project=>nil, :skip=>nil, :limit=>nil, :flags=>[:slave_ok]}, @project=nil, @skip=0, @limit=0, @flags=[:slave_ok]>   
```

Then I've found `#payload` and thought: hey, that's the proper way! let's commit:
But there's not even the collection =/

```
{:command_name=>"find", :database_name=>"db", :command=>{:filter=>{"_id"=>"xxxxxx"}, :project=>nil, :skip=>nil, :limit=>nil, :flags=>[:slave_ok]}, :request_id=>nil}   
```


Fixing a better inspect for this. Pushing today.
Sorry and thanks.

nofxx:
Here's the `#payload` method:

https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/protocol/query.rb#L80

Even the comment says it's meant for debugging/monitoring. Did I get it wrong?
Not sure if Ishould I file a bug reporting it doesn't contains the collection name.

nofxx:
Durran helped me on JIRA: 

https://jira.mongodb.org/browse/RUBY-1000

Implementing the new way.

coin8086:
@nofxx I encountered a problem on the payload method you mentioned:

2016-03-23 23:24:26 - NoMethodError - undefined method `payload' for #<Mongo::Protocol::Query:0x00000003356cb0>:
	/home/robert/.rvm/gems/ruby-2.2.1/gems/rack-mini-profiler-0.9.9.2/lib/patches/db/mongo.rb:6:in `dispatch_with_timing'

It's in my Sinatra App with Mongo and Thin. When I turn off the profiler it's all right. It seems that the payload method is longer there. My MongoDB version is 3.0 and Ruby Mongo Driver is 2.0. Do I miss/mistake anything?

PR #157, opened 2015-08-19
Support IPv6 addr on Windows Platform.

nazy:
Replace ':' of key with '_' at FileStore#path method to avoid to include
unusable path charactor on Windows.
When using IPv6, localhost will be '::1' and it leads the "MiniProfiler
storage failure: Invalid argument @ rb_sysopen - #{path}" error.

### DIFF
--- a/lib/mini_profiler/storage/file_store.rb
+++ b/lib/mini_profiler/storage/file_store.rb
@@ -26,8 +26,14 @@ module Rack
         end
 
         private
-        def path(key)
-          @path + "/" + @prefix  + "_" + key
+        if RUBY_PLATFORM =~ /mswin(?!ce)|mingw|cygwin|bccwin/
+          def path(key)
+            @path + "/" + @prefix  + "_" + key.gsub(/:/, '_')
+          end
+        else
+          def path(key)
+            @path + "/" + @prefix  + "_" + key
+          end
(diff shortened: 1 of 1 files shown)


SamSaffron:
looks good

charlesdeb:
Hmm, it's now 2021, and I have version 2.3.2 of this gem in my rails 6.1 app, running on Ubuntu 20.04 - although my application code is living on an NTFS partition (Win 10 Pro). I am getting the same error message when visiting pages in my app:

```
MiniProfiler storage failure: Invalid argument @ rb_sysopen - /media/edrive/path/to/my/app/tmp/miniprofiler/mp_views_::1
```
I am happy to provide further system info if you let me know what you need to see. For now, I have just commented this out of my gem file - but I expect I'll want to use this gem in the future.

SamSaffron:
I would certainly love this to work, simplest workaround would be using Redis here. 

Are you using Ruby on Windows or Ruby via WSL2?

charlesdeb:
I am using ruby 3 via rbenv running directly on Ubuntu 20.04. However, the
files in my project live on an NTFS file system that is mounted within
Ubuntu. I am not running ruby on windows nor using any version of WSL.

How would I configure redis?

PR #158, opened 2015-08-25
Removed unused statement

printercu:
`authorization_mode` is set later in same method.


kbrock:
confused about title - this is deleting `skip_schema_queries = true` not `authorization_mode`

printercu:
Yeah, had a mistake. There is `c.skip_schema_queries = (Rails.env.development? || Rails.env.test?)` later in method.

SamSaffron:
looks like we already sorted this, thanks!

PR #159, opened 2015-09-14
Updating README to include MRI 2.2 in flamegraph generation.

leighhalliday:
It seems to work fine with MRI 2.2 so I wanted to update the README file to include this version.


SamSaffron:
good point perhaps 2.0+ ?

PR #161, opened 2015-10-09
README header was not clear enough regarding authorization

jcblitz:
It is not clear from the README that non-development environments need to authorize, not just production. If you read the comments in the source, it refers to 'non-development' as well.


SamSaffron:
sure, thanks!

PR #162, opened 2015-10-12
allow travis to use container-based infrastructure

dgynn:
This is just some minor housekeeping to modernize the Travis config. It also explicitly sets ruby versions for 2.0 and 2.1 to latest version so tests don't run against 2.0.0 and 2.1.0.


SamSaffron:
sure sounds good!

PR #166, opened 2015-10-13
Add support for Oracle through the activerecord-oracle_enhanced-adapter

rrooding:


SamSaffron:
pretty concerned with this one, can we guard the patch so it only applies to tested AR versions?

rrooding:
I've added a check to only load for oracle-enhanced ~1.5. It should also work for ~1.6 because there haven't been any changes in that part of the gem, but I can't confirm that yet, will create a pull later if necessary.

SamSaffron:
cool, lets try it out then.

PR #167, opened 2015-10-13
Don't block the mongo patch when other DB patches are already loaded

rrooding:
Found this when fixing #166. I think only ActiveRecord and Sequel should be blocked when other patches are already loaded because they would actually conflict. This check was probably added by accident.


kbrock:
:+1: on this one. This should not be in there.

@rrooding if you don't rebase, I can do it as well.

kbrock:
I want to change riak as well, I'll just rebase it

PR #172, opened 2015-11-16
Add support for NoBrainer (rethinkdb) profiling

niv:
This is a simple profiler integration that exposes the builtin NoBrainer
profiler to rack-mini-profiler. Should be pretty straightforward.

Highlighting is a bit off since mrp expects SQL in the frontend, but
nothing too shabby.

Also embedding some metadata in the profiler query string (things like
index usage) that the developer probably wants to see, but mrp hasn't
extra fields for.

--

Let me know what you think! :)

screen shot 2015-11-16 at 16 03 36


SamSaffron:
awesome, thanks!

PR #173, opened 2015-11-18
Set X-MiniProfiler-Ids when status is 2xx

tyamagu2:
This makes profiling API that uses status code other than 200 for successful response a little easier.

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -291,7 +291,7 @@ module Rack
       if (config.authorization_mode == :whitelist && !MiniProfiler.request_authorized?)
         # this is non-obvious, don't kill the profiling cookie on errors or short requests
         # this ensures that stuff that never reaches the rails stack does not kill profiling
-        if status == 200 && ((Time.now - start) > 0.1)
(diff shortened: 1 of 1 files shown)


lib/mini_profiler/profiler.rb
```diff
@@ -336,7 +336,7 @@ def call(env)
         @storage.save(page_struct)
 
         # inject headers, script
-        if headers['Content-Type'] && status == 200
+        if status >= 200 && status < 300
```
tyamagu2: I removed `headers['Content-Type']` from the condition to set X-MiniProfiler-Ids when status is 204 (No Content). I believe this is not a problem since X-MiniProfiler-Ids was set regardless of `headers['Content-Type']` until c101e7463d6ebf6d3b0078e81e9a7b7ddd17fa18. `headers['Content-Type']` is still used to decide whether to inject js to body in inject_profiler, just like c101e7463d6ebf6d3b0078e81e9a7b7ddd17fa18 did.

SamSaffron:
ok lets go with this, thank! 

I did notice we are double setting the cookie in some cases, maybe worth debugging into it.

tyamagu2:
Thanks. I'll look into that problem when I have time to.

PR #174, opened 2015-11-19
Prevent ENV dumps via new config option

mathias:
I had a concern about dumping ENV vars into HTML in an app, and I wanted to disable it completely, so I added a config option here that just prevents `?pp=env` from functioning.

### DIFF
--- a/lib/mini_profiler/config.rb
+++ b/lib/mini_profiler/config.rb
@@ -13,7 +13,7 @@ module Rack
     end
 
     attr_accessor :authorization_mode, :auto_inject, :backtrace_ignores, :backtrace_includes, :backtrace_remove,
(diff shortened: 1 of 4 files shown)


lib/mini_profiler/config.rb
```diff
@@ -13,7 +13,7 @@ def self.attributes
     end
 
     attr_accessor :authorization_mode, :auto_inject, :backtrace_ignores, :backtrace_includes, :backtrace_remove,
```
mathias: Should I also reformat this list of attrs to be a certain max width per line, too? (80 cols or 100 cols?) Or leave it?

SamSaffron: sure, feel free to clean that up

SamSaffron:
sure, sound fine, thanks!

mathias:
Thank you!

PR #175, opened 2015-11-20
Re-indent at 80 cols

mathias:
As promised in #174, here's the config options attrs list re-indented at 80 columns width.


SamSaffron:
Thanks !

mathias:
Great, thanks!

Would you mind cutting a minor version release with this and #174 in it? We'd like to point at a version on rubygems in our project.

Thanks!!!

SamSaffron:
Cutting one right now!

PR #177, opened 2015-11-30
Remove unused vars from profiling class

kbrock:
Noticed a few variables not being used.

If you prefer these as separate PRs, or as a single commit, just give the word

Thanks for the great tool.


SamSaffron:
looks good!

PR #178, opened 2015-12-01
lint wants x !== null or x === null

kbrock:
I'm just cleaning up a few errors from lint in sublime.
If you have a reason for keeping this as `!= null`, please just close this PR.

Thanks again for the great tool.


SamSaffron:
sure ... looks good

PR #181, opened 2015-12-26
Fix jQuery templates not being fetched when jQuery dataType is defaulted to JSON

bradleypriest:
Explicitly specify a text response type for `jQuery.get` in case the application has set the default via 
`$.ajaxSetup({dataType: 'json'})` or similar.

See the last example on the docs page. https://api.jquery.com/jquery.get/

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -57,7 +57,7 @@ var MiniProfiler = (function () {
                     $('body').append(data);
                     success();
                 }
-            });
+            }, "text");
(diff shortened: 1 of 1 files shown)


bradleypriest:
Any concerns with this?

seanlinsley:
This seems to make sense, cc @SamSaffron

SamSaffron:
sorry for the delay here, I think it seems fine!

PR #182, opened 2015-12-27
Prevent intermittent spec failures on profiler_spec

dgynn:
This spec modifies `Time.now` but not before time has ticked.


SamSaffron:
Nice one! Thanks :heart:

PR #183, opened 2015-12-27
Support Rack 2.0 for Rails 5

dgynn:
The `Rack::File` API has changed again with Rack 2.0, which is required for Rails 5. The profiler uses Rack::File to serve the CSS and JS files. Everything else seems to work fine with Rails 5.

This PR changes the way Rack::File is used to work with both Rack 1.2 and 2.0. This *does* break compatibility with Rack 1.1. My assumption was that anyone profiling an app would no longer be on a super old version of Rack.

Let me know if you'd rather keep Rack 1.1.x support and I can modify the code.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rack-mini-profiler (0.9.8)
-      rack (>= 1.1.3)
+      rack (>= 1.2.0)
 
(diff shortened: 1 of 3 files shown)


SamSaffron:
More than happy to drop Rack 1.1, can you update the changelog with that, it is ancient and people can simply install an older gem if they must

PR #184, opened 2015-12-27
Performance speedups for gc-profiler

dgynn:
This PR tunes up the way the `gc-profiler` code analyzes the results. The speed improvement depends on the number of objects in memory. In rough measurements this is about 3x faster with times decreasing from ~11 seconds to ~3.5 seconds.

I've also reformatted the output to make it easier to do before/after comparisons.


SamSaffron:
awesome, really happy to have you looking at this stuff :heart:

Issue #185, opened 2016-01-06
ArgumentError: comparison of String with 200 failed


paweljw:
Hi,

This error in the title happens on every request. Despite that, MiniProfiler seems to work and produce meaningful results. Also this doesn't happen in the development environment, only in staging (and production, when MiniProfiler is configured to run there).

Here's a complete stack trace of an example error:

```
[2016-01-06 11:48:39] ERROR ArgumentError: comparison of String with 200 failed
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/rack-mini-profiler-0.9.8/lib/mini_profiler/profiler.rb:294:in `>='
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/rack-mini-profiler-0.9.8/lib/mini_profiler/profiler.rb:294:in `call'
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/railties-4.2.0/lib/rails/engine.rb:518:in `call'
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/railties-4.2.0/lib/rails/application.rb:164:in `call'
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/rack-1.6.4/lib/rack/lock.rb:17:in `call'
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/rack-1.6.4/lib/rack/content_length.rb:15:in `call'
  /home/someuser/.rvm/gems/ruby-2.2.1@rails4.2/gems/rack-1.6.4/lib/rack/handler/webrick.rb:88:in `service'
  /home/someuser/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/webrick/httpserver.rb:138:in `service'
  /home/someuser/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/webrick/httpserver.rb:94:in `run'
  /home/someuser/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/webrick/server.rb:294:in `block in start_thread'
```

Reading the code in profiler.rb, it seems that the offending line is

```ruby
if status >= 200 && status < 300 && ((Time.now - start) > 0.1)
```

so obviously I tried the most obvious fix I could: I forked rack-mini-profiler and forced a cast to integer there, so that the line reads

```ruby
[...]

PR #186, opened 2016-01-07
Fix ArgumentError: comparison of String with 200 failed

paweljw:
Issue #185 - converts status given as string (from badly-behaved Rack middlewares) to integer before comparison.


SamSaffron:
thanks

PR #187, opened 2016-01-07
Allow backtrace_includes to work with an array of regexes

dgynn:
With 2 Regexps configured in backtrace_ignores, no stack is ever shown. For example... 
```ruby
Rack::MiniProfiler.config.backtrace_includes = [/^\/?(app|config|lib|test)/] + [/my_special_gem/]
```
This PR changes `all?` to `any?` to match the backtrace_includes Regexps. 
The change to `none?` on the backtrace_ignores is just for readability.


SamSaffron:
good catch! thanks

PR #188, opened 2016-01-08
Try to use localStorage to avoid errors in Safari

elia:
Safari private windows will have localStorage available but with no usable quota.
In that scenario localStorage should be avoided.

Further explanation here: http://stackoverflow.com/a/14555361


SamSaffron:
sure, this should work, I wonder if we should memoize though.

SamSaffron:
thanks !

PR #190, opened 2016-01-18
allow class level methods to be profiled

kbrock:
I would like to profile a class level method.

Let me know if you would like a different approach.

---

If a method is to be profiled and it is not an instance method, then
define it on the singleton instance

This will only allow people to profile an instance method when both
a class level and instance level have the same name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 
+
+- [ADDED] Rack::MiniProfiler.profile_singleton_method (@kbrock)
+
 ## 0.9.8 - 2015-11-27 (Sam Saffron)
 
 - [FEATURE] disable_env_dump config setting (@mathias)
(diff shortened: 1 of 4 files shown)


SamSaffron:
I am not sure, I think we need some sort of extra option or method name for it, cause #foo can be defined both at the class and instance level.

kbrock:
@SamSaffron that method has a bit of guts in it, you want another define that simply calls the original and passes in the `singleton`?

SamSaffron:
Im thinking maybe a special method for it

`profile_singleton` or perhaps add an extra param to profile_method... key
is that I want you to be able to profile:

```
class Foo
   def bar
   end

   def self.bar
   end
end
```

kbrock:
Thanks @SamSaffron 

This should work better for that scenario.
Is this closer to what you had in mind?

SamSaffron:
Perfect! Can you add in some documentation for this stuff in the readme and a note into the changelog?

kbrock:
hmm. I felt the method name was too kurt, but the description I chose basically just used the same words :/ do you like `[feature]` or just plain `[add]`?

SamSaffron:
whatever works for you :)

SamSaffron:
thanks heaps!

PR #191, opened 2016-01-19
Put options into a table

kbrock:
Hello @SamSaffron 

I hope you like the more active voice, table formatting, and `code` formatting.

If you like just one or the other, just let me know and I can fix

You can view the final result [here](https://github.com/kbrock/rack-mini-profiler/blob/reformat_readme_options/README.md#configuration-options).

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,11 +145,11 @@ if Rails.env.production?
 end
 ```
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
sure, looks good to me!

PR #192, opened 2016-01-19
make rack example complete in readme

### LINKED ISSUE #176: rack-mini-profiler badge not showing in rack app?
I was having a problem getting rack-mini-profiler working so I created a trivial rack app based on an example I saw

```
require 'rack'
require 'rack-mini-profiler'

use Rack::MiniProfiler
run Proc.new { |env| ['200', {'Content-Type' => 'text/html'}, [ "Hello!" ]] }

```

In loading this up, it seems like it does not work.. no badge, no nothing.  I finally tried ?pp=enable and I was then able to use ?pp=help at least, but still no badge of any kind showing.  

How do you get the badge to show in a rack app!?!?

kbrock:
fixed #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,17 @@ end
 
 ```ruby
 require 'rack-mini-profiler'
+
+home = lambda { |env|
+  [200, {'Content-Type' => 'text/html'}, ["<html><body>hello!</body></html>"]]
+}
+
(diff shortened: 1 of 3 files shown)


SamSaffron:
Thanks!

PR #193, opened 2016-01-20
Allow hosts to be skipped from profiling

taoza:
Some applications may include sub-apps with a different hosts for better isolation. If we allow paths to be skipped, I guess it also make sense to skip hosts.


SamSaffron:
hmmm this feels a bit odd to me, can this not be done in any other way? simply don't authorize the host?

taoza:
@SamSaffron Thanks for pointing this out. Yup, it make sense to just use the `pre_authorize_cb` to get the host from rack env. The proposed solution is more for the convenience, especially since we already have the filters on the paths. But I'm happy to close this one if it doesn't worth it.

PR #197, opened 2016-02-01
Generate html method

kbrock:
Want to generate the html from the command line.
Extracted the actual creation of the html in a separate method.

Sorry, got a little carried away modifying `generate_html` method.
Also trimmed down `insert`.
Want to call `create_context` and return the context. It seems odd to set the thread local variable so many times. I didn't find any places in the initialization chain that access `current` (or would need to in the guessable future)

Let me know if you want these as separate PRs (or if I should just get rid of any of this)

Thanks,
Keenan

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -388,39 +388,17 @@ module Rack
     end
 
     def inject(fragment, script)
-      if fragment.match(/<\/body>/i)
-        # explicit </body>
-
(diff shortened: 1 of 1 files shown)


SamSaffron:
In principle I am fine with the simplification here but specs are failing, something is off.

kbrock:
@SamSaffron thanks. not sure how the `s` got in there.
Working now

SamSaffron:
thanks!

PR #198, opened 2016-02-01
Added a patch for neo4j OGM.

ProGM:
In this period I'm using Neo4j using this Object Graph Mapper:

https://github.com/neo4jrb/neo4j

I created a patch to profile cypher queries for rack-mini-profiler. Since it's working, I'm sharing it ;)

Example of output:
schermata 2016-02-01 alle 17 54 20

### DIFF
--- a/lib/patches/db/neo4j.rb
+++ b/lib/patches/db/neo4j.rb
@@ -1,0 +1,14 @@
+class Neo4j::Core::Query
+  alias_method :response_without_miniprofiler, :response
+
+  def response
+    return @response if @response
+    start = Time.now
+    rval = response_without_miniprofiler
+    elapsed_time = SqlPatches.elapsed_time(start)
+    Rack::MiniProfiler.record_sql(to_cypher, elapsed_time)
+    rval
+  end
+
+  alias_method :response_with_miniprofiler, :response
+end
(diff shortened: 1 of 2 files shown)


SamSaffron:
cool, looks good!

cheerfulstoic:
This is great, thanks!

I would check out the `Neo4j::Server::CypherSession.log_with` method.  It allows you to give a block that will subscribe to queries using ActiveSupport`s Notification service.

If you want to make it future proof you might also check out `Neo4j::Core::CypherSession::Adaptors::Base.subscribe_to_query` which is the way it will be done in version 7.0 of the `neo4j-core` gem (which `neo4j` uses).  You can see them used here:

https://github.com/neo4jrb/neo4j-core/blob/master/spec/spec_helper.rb#L58

Actually, now that I think about it that is only going to give you a formatted string.  You should probably subscribe to the events yourself like:

```ruby
ActiveSupport::Notifications.subscribe('neo4j.cypher_query') 
```

That will give you extra data which will, among other things, let you figure out how long the query took.

Also, keep in mind that in 7.0 you'll be able to make multiple queries in the same request.  So accordingly there is an instrument for each query and instruments for each request.

Hopefully that all makes sense and you'll be able to trace down everything from the `spec_helper.rb` link above, but if not let me know!

Oh, also, in the new code note that I made an `Instrumentable` module, so searching for the method names might not be as easy because they're generated from that:

https://github.com/neo4jrb/neo4j-core/blob/master/lib/neo4j/core/instrumentable.rb

PR #200, opened 2016-02-02
Highlight each result as they appear in UI

ericboehs:
This is a port of https://github.com/MiniProfiler/ui/pull/25. I wasn't sure which repository I was supposed to submit to.

In action:

screen capture on 2016-02-02 at 15-59-27


kbrock:
nice

SamSaffron:
This kind of feature is going to have to be optional opt in, if it kept doing that as I was browsing discourse it would be too much, also css changes need to go in sass and then sass needs to be compiled.

PR #201, opened 2016-02-03
More robust checking for serves_static_files?

ianks:
Rails 5 switched the configuration option for serving static files again. Lets check it.

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -24,7 +24,12 @@ module Rack::MiniProfilerRails
 
     c.skip_paths ||= []
 
-    c.skip_paths << app.config.assets.prefix if serves_static_assets?(app)
+    begin
+      c.skip_paths << app.config.assets.prefix if serves_static_assets?(app)
+    rescue
+      # app.config did not repond_to? :assets
+    end
+
 
     if Rails.env.development?
       c.skip_schema_queries = true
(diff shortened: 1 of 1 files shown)


lib/mini_profiler_rails/railtie.rb
```diff
@@ -24,7 +24,12 @@ def self.initialize!(app)
 
     c.skip_paths ||= []
 
-    c.skip_paths << app.config.assets.prefix if serves_static_assets?(app)
+    begin
+      c.skip_paths << app.config.assets.prefix if serves_static_assets?(app)
+    rescue
+      # app.config did not repond_to? :assets
```
SamSaffron: This feels odd its hiding a possible error + first line is !app.respond_to(:assets)

ianks: > first line is !app.respond_to(:assets)

I noticed this as well, and was somewhat confused by it. Notice here we are checking if app.*config* responds to :assets, not just `app`. What is the reason for this?

SamSaffron: I say we just fix that then, I am uneasy about hiding the error in a global
rescue

ianks:
I removed to global rescue and made sure we check `app.config` instead of `app.assets`. I'm not entirely sure where the code `app.assets` originated from. Should we add another check which also checks `app.assets`? Or do we conclude it was a bug to begin with?

SamSaffron:
not sure why it is there, maybe a bug or a check that asset pipeline is enabled? Happy to pull this in though.

PR #202, opened 2016-02-09
Capture parameters for sql

kbrock:
Hi Sam/All,

Looks like the query parameters are not captured.
Not sure if this is by design, or an oversight.

Would you like a configuration parameter to turn this on or off.
Or some extra protections to limit passwords or large values?

Thanks
Keenan

### DIFF
--- a/lib/mini_profiler/config.rb
+++ b/lib/mini_profiler/config.rb
@@ -18,6 +18,7 @@ module Rack
       :flamegraph_sample_rate, :logger, :pre_authorize_cb, :skip_paths,
       :skip_schema_queries, :storage, :storage_failure, :storage_instance,
       :storage_options, :user_provider
+    attr_accessor :skip_sql_param_names, :max_sql_param_length
 
     # ui accessors
     attr_accessor :collapse_results, :max_traces_to_show, :position,
@@ -49,6 +50,8 @@ module Rack
           end
           @enabled = true
           @disable_env_dump = false
+          @max_sql_param_length = 0 # disable sql parameter collection by default
+          @skip_sql_param_names = /password/ # skips parameters with the name password by default
 
           # ui parameters
           @autorized          = true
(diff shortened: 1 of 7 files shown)


SamSaffron:
there is a reason for this, imagine you are inserting a huge file or bit of text, ideally we should truncate params so we don't store giant blobs in redis.

kbrock:
I can add tests, but is this closer to what you envisioned?

SamSaffron:
In principle I am ok with this, but we need to work out the UI and tests need to be added

We also need to confirm there is no noticeable perf impact.

kbrock:
Any suggestions on measuring the overhead that is introduced?
On my computer, benchmarks vary so much for ruby only micro tests. Not sure if I could detect issues introduced with an integration test.

If I pulled off the first commit into a separate PR, could we merge that? It would not do anything for users. But it makes my local fork just a little closer to the master version.

I hadn't looked into implementing this in the ui. Locally, I'm using a ruby ui to access the data. It is easier for me to run a method a number of times and then generate a single table with correlated results.

SamSaffron:
Lets make this opt-in default off, then I am find to merge it in.

kbrock:
Hi Sam

I made this default to disabled.
Tried to make the placement of the `params` parameter the most logical, but it was tricky.

Is this closer to what you had in mind?

--Keenan

SamSaffron:
Yes, this looks good, can you also add some doco into the readme?

pgiblock:
Is this feature supposed to work? I noticed it is an undocumented feature. I set:
```
Rack::MiniProfiler.config.max_sql_param_length == 100
```
but unfortunately, I do not see any parameter values (and the JSON of the each result has `parameters=null`)

kbrock:
The data stored in redis has the parameters. Not sure how it translates to the rest of the system

PR #204, opened 2016-02-16
fix get_profile_script when running under passenger configured with RailsBaseURI

nspring:
This PR addresses a problem where running a Rails application under Passenger, which has been configured using its RailsBaseURI configuration option.  The mini-profiler would fail to load/display because the generated URL would lack the configured Base.  Other features would work, for example, if given the pp= option.  

The problem appeared to be caused by the fixes to #25 and #73 that handle rails engines by discarding SCRIPT_NAME when creating the URL to be injected.  This fix uses the environment variable PASSENGER_BASE_URI to act as SCRIPT_NAME had before the fix.  Rails engines are unlikely to alter the Passenger-specific variable as they altered SCRIPT_NAME, and any broader change might interact poorly with rails engines.  This patch may work if both passenger *and* rails engines are in use, but that is untested.


SamSaffron:
ok ... lets try this

PR #207, opened 2016-02-20
Enable Riak and Mongo with PG

### LINKED ISSUE #194: Mongo queries don't show up when PG gem is also loaded
Hello!

Thanks for this wonderful gem.

I saw some possibly related discussion in https://github.com/MiniProfiler/rack-mini-profiler/issues/180, but my issue is a bit different.

In my app, I have both the "pg" gem and the "mongoid" gem loaded. Because of the order of the requires [here](https://github.com/MiniProfiler/rack-mini-profiler/blob/cb7a6cf4d4cd7febb587bcffb4719233034c331b/lib/patches/sql_patches.rb#L49), the PG patch is loaded, and then the Mongo patch is skipped, because rack-mini-profiler doesn't patch more than one db (if I'm understanding this right).

For my app, Mongoid is the main db, and postgres is used in one minor part of the app, so this behavior isn't ideal for me. I've gotten around it for now by lazily requiring PG only in the one file that needs it, but I was a bit confused for a few minutes, and wanted to share the scenario in case it can help inform any decisions about how to specify the preferred patch.

Cheers and thanks again :smile:

kbrock:
This allows Riak and Mongo to work with a relational database.

The `SqlPatches.patched?` is only there to tell active_record/Sequel that the drivers are patched an there is no need to also patch active record.

FYI: you need to `gem "pg"` before `rack-mini-profiler` to use the native driver version.

fixes #167 (as it cherry picks the code)
fixes #194


SamSaffron:
good point! thanks!

PR #208, opened 2016-02-20
Create tmp directory when actually using it

### LINKED ISSUE #152: Runs mkdir in tmp even when not using FileStore
Is it necessary to do this when FileStore isn't configured?
https://github.com/MiniProfiler/rack-mini-profiler/blob/f4d12ae8b621c7bd7762bcd67208ff964858f777/lib/mini_profiler_rails/railtie.rb#L43

In our production servers the app doesn't have permissions under tmp/, so we get an error on startup. I wonder if this would be an easy fix. Awesome tool, can't wait to start seeing performance stats. :+1:

kbrock:
This still creates a directory for filestore before being used, since
we don't want to do this lazy and possibly introduce a race condition.

But it now waits later in the configuration process, when profiler
creates the storage_instance

Fixes #152

### DIFF
--- a/lib/mini_profiler/storage/file_store.rb
+++ b/lib/mini_profiler/storage/file_store.rb
@@ -44,6 +44,8 @@ module Rack
         @path = args[:path]
         @expires_in_seconds = args[:expires_in] || EXPIRES_IN_SECONDS
         raise ArgumentError.new :path unless @path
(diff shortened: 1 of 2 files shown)


SamSaffron:
true, also works better outside of rails

PR #209, opened 2016-02-20
throw away over 10 unviewed ids

### LINKED ISSUE #127: Profiler shows old data from previous requests
The profiler keeps showing some "stuck" data from old requests on every request - i.e. I have a few rows persisting with the same value and an older timestamp, regardless if this request has been made at all. The actual current request data is still appearing, but with the old one mixed in in between. I have tried restarting the Rails server, restarting my browser and clearing its cache. I haven't seen any other references to this issue and it's not described as a feature. What can I do? Has anyone else had the same problem? Am I missing something?

kbrock:
fixes #127 

if there are more than 10 unviewed ids, it actually removes them from the storage.

Not thrilled about adding a method to Store.
@SamSaffron would you prefer just removing the `[0..8]`?

I didnt see how this would ever mark these as viewed.

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -552,8 +552,12 @@ Append the following to your query string:
     end
 
     def ids(env)
-      # cap at 10 ids, otherwise there is a chance you can blow the header
-      ([current.page_struct[:id]] + (@storage.get_unviewed_ids(user(env)) || [])[0..8]).uniq
+      all = ([current.page_struct[:id]] + (@storage.get_unviewed_ids(user(env)) || [])).uniq
+      if all.size > @config.max_traces_to_show
+        all = all[0...@config.max_traces_to_show]
+        @storage.set_all_unviewed(user(env), all)
+      end
+      all
     end
 
     def ids_json(env)
--- a/lib/mini_profiler/storage/abstract_store.rb
+++ b/lib/mini_profiler/storage/abstract_store.rb
@@ -18,6 +18,10 @@ module Rack
         raise NotImplementedError.new("set_viewed is not implemented")
       end
 
+      def set_all_unviewed(user, ids)
+        raise NotImplementedError.new("set_all_unviewed is not implemented")
+      end
+
       def get_unviewed_ids(user)
         raise NotImplementedError.new("get_unviewed_ids is not implemented")
       end
--- a/lib/mini_profiler/storage/file_store.rb
+++ b/lib/mini_profiler/storage/file_store.rb
@@ -114,6 +114,12 @@ module Rack
         }
       end
 
+      def set_all_unviewed(user, ids)
+        @user_view_lock.synchronize {
+          @user_view_cache[user] = ids.uniq
+        }
+      end
+
       def get_unviewed_ids(user)
         @user_view_lock.synchronize {
           @user_view_cache[user]
--- a/lib/mini_profiler/storage/memcache_store.rb
+++ b/lib/mini_profiler/storage/memcache_store.rb
@@ -43,6 +43,10 @@ module Rack
         end
       end
 
+      def set_all_unviewed(user, ids)
+        @client.set("#{@prefix}-#{user}-v", ids, @expires_in_seconds)
+      end
+
       def get_unviewed_ids(user)
         @client.get("#{@prefix}-#{user}-v") || []
       end
--- a/lib/mini_profiler/storage/memory_store.rb
+++ b/lib/mini_profiler/storage/memory_store.rb
@@ -98,6 +98,12 @@ module Rack
         }
       end
 
+      def set_all_unviewed(user, ids)
+        @user_view_lock.synchronize {
+          @user_view_cache[user] = ids
+        }
+      end
+
       def get_unviewed_ids(user)
         @user_view_lock.synchronize {
           @user_view_cache[user]
--- a/lib/mini_profiler/storage/redis_store.rb
+++ b/lib/mini_profiler/storage/redis_store.rb
@@ -33,6 +33,13 @@ module Rack
         redis.expire key, @expires_in_seconds
       end
 
+      def set_all_unviewed(user, ids)
+        key = "#{@prefix}-#{user}-v"
+        redis.del key
+        ids.each { |id| redis.sadd(key, id) }
+        redis.expire key, @expires_in_seconds
+      end
+
       def set_viewed(user, id)
         redis.srem "#{@prefix}-#{user}-v", id
       end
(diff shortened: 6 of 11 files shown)


SamSaffron:
This has been bugging me forever!!! 

Can you add some tests here, then we can merge it in

SamSaffron:
@kbrock I want to pull this in, but we got to have at least one test here... also backlog is more configurable now so it should use the config setting ...

lib/mini_profiler/profiler.rb
```diff
@@ -553,7 +553,12 @@ def flamegraph(graph)
 
     def ids(env)
       # cap at 10 ids, otherwise there is a chance you can blow the header
-      ([current.page_struct[:id]] + (@storage.get_unviewed_ids(user(env)) || [])[0..8]).uniq
+      all = ([current.page_struct[:id]] + (@storage.get_unviewed_ids(user(env)) || [])).uniq
+      if all.size > 10
+        all = all[0...10]
+        @storage.set_all_unviewed_ids(user(env), all)
```
stiller-leser: This does not work and has to be renamed to `@storage.set_all_unviewed(user(env), all)`

kbrock:
@SamSaffron I must be looking in the wrong spot. This is a javascript configuration value?

You want me to setup a spot to put it into the ruby config?
You want me to pass that value down to the "get comma ids"? (not necessary if it is in ruby config)

SamSaffron:
Its called max_traces_to_show, now defaults to 20

kbrock:
thanks

- using configuration params
- also added specs for all storages

kbrock:
@SamSaffron is this closer to what you had in mind?

SamSaffron:
Let's try this, it looks good to me

chmurph2:
Did this PR actually fix the original issue, #127? In particular:

>The profiler keeps showing some "stuck" data from old requests on every request - i.e. I have a few rows persisting with the same value and an older timestamp, regardless if this request has been made at all. 

I bundled this PR and I still see the same "stuck" data issue:

journalservice 2016-03-27 08-22-45

which expands to:

journalservice 2016-03-27 08-23-18

Seems like a different issue was resolved.

chmurph2:
To clarify, when I first load the page in the previous comment, there's only one request displayed by the profiler.  As I continue to refresh the page, it builds up these historical requests, and the top one isn't always the most recent.

kbrock:
@chmurph2 Thanks for reporting this.
I have seen this here and there, but was unable to reproduce locally, so this was more of a blind solution.

Would you be willing to put some debugging information [here](https://github.com/MiniProfiler/rack-mini-profiler/pull/209/files#diff-810de93622d93c35de753db1b046abc8R555)?

1. What datasource are you using?
2. Can you tell if it is detecting the issue?
3. Can you tell if it is detecting the correct `max_traces_to_show`?
3. Can you tell if it is calling into the (correct) datasource to clear the ids out?

Thanks for any pointers you can share

chmurph2:
@kbrock:

Here's the debugging information at that line, when I'm seeing `> 20` ids/traces:

```ruby
["w2fjm7roo2ozi27vben", "8iyzy3mgjwrlud9oo5tr", "o1isja1dtar6fxqim2en", "vfdxbe3kcznt14qeus5w", "x44bcgg0xt77lacyxp0d", "43dl0g1o3i231ql7r7ae", "a37pt4i82mfbvmu57gzy", "wg005updjynanef7a3b2", "q1c64hap5ytqepwxsi25", "ct04inxfh4bc0e0prexv", "ug2aunsgwnvvvfm14v2o", "ljtnugabaj40yj53ob4a", "utp6341ogvu1f3eswxqz", "4nbe2ztnxz7f4h1brs8d", "j26bksw5p2yghzabu594", "v5m0tciodfe2g5q5gquo", "tivqikoj57tkhcsfdku2", "p7yx7av3phswvrw2spia", "gbh57j6wvt7hktrmhj53", "45vzjwojnw5gndt2j504", "5iv60gq640urql75l0kq"]
```

1. I assume you mean the "storage option". I am using the default: FileStore.
2. When max traces is `> 20` it is entering [this code path](https://github.com/MiniProfiler/rack-mini-profiler/pull/209/files#diff-810de93622d93c35de753db1b046abc8R557). But see 4 below.
3. `@config.max_traces_to_show` is defaulting to 20. 
4. It seems to at least call it: `<Rack::MiniProfiler::FileStore:0x007f9ca544f5e0>`. The id array never goes above 21 once I hit the max traces.

So, I guess it's setting anything `> 20` unviewed?  I'm not sure I'm groking the intention, though.  Why would I want to see `ids` 1..19 displayed, regardless?

Anyway, I hope this helps in some fashion.

kbrock:
@chmurph2 thanks. So even though you're setting `max_traces_to_show` to a lower value, it is not picking it up.

How are you setting `max_traces_to_show` and where?

chmurph2:
No, for the above test I let `max_traces_to_show` default to 20.  I'm not setting it now.

chmurph2:
When I did set `max_traces_to_show` to 1 before the above test, it would only display one id/trace per page view.  But I wasn't sure if this was what I should be doing or not... seems like if there are legitimate multiple requests in the page, I would only see one of them?

kbrock:
There was an issue that if this list of traces grew too large, then when it sent down the ids of the traces to the client, it would blow up  the http headers.
So a threshold was introduced. It was hardcoded to 10 until `max_traces_to_show` was recently introduced).
The bug this PR was focused is once that list of trace ids went above the threshold, the trace ids would never be cleared out. This PR was a little heavy handed but said, we're over our threshold, just assume that the client doesn't want to see these old traces anymore.

@SamSaffron What does "viewed" mean? a) the id was sent to the client b) the client clicked on that particular trace? c) the browser sent an "I've rendered this page" response to the server and client timings were put into the trace?

if it is a, then should this code just clear out all of the trace ids? The threshold would just limit the ids sent back.
If it is c, then maybe we need to focus on ajax requests and mark them as viewed?

SamSaffron:
It is a, we got to consider stuff viewed as soon as it is sent to the
client, otherwise stuff gets weird and you carry a huge backlog.

chmurph2:
Please let me know if you need me to test anything else for this.

kbrock:
@chmurph2 maybe we can open a new issue instead of commenting on this old one?

This was focused at capping the events so we at least deleted the extra ones.

But it sounds like you have a different problem. Old events seem to be just hanging around.

chmurph2:
@kbrock OK, I can open a new issue.  I want to note #127 details the problem I'm still having, though that one is now closed.  Should that one be reopened instead?

PR #210, opened 2016-02-20
add accessors to page, request timer struct

kbrock:
I'm doing more stuff in ruby and find these 2 methods helpful

I could move into `TimerStruct::Base` if you prefer.


SamSaffron:
sure, seems fine

Issue #199, opened 2016-02-01
Using ActiveSupport::Notifications


kbrock:
Sam / all,

Just saw an [rack-profiler](https://github.com/dawanda/rack-profiler) which uses notifications and seems to capture much of what is interesting.

Other than the effort required, are there obvious issues in transitioning across?
(aka - I'm not sure if this is a feature that is much harder than it looks at the surface or if there are parts missing from notifications that prevent transition across)

Thanks,
Keenan

SamSaffron:
I have no objections to adding an option for sipping the notifications
firehose, but there are issues

1. It is less accurate than the current patches (which split time reading
records and time executing sql)
2. It add more overhead than current profiling
3. It is way too noisy to be default on, render 100 partials in a page and
disaster pasta
4. It is less compatible (we work fine without active support)

Open to adding an icon or pp option to enable the notification firehose.

kbrock:
@SamSaffron thanks you for your insight

I was having trouble getting the SQL tracing working with a rails 3.2 version of our app. So I had thought just tapping into notifications would be simpler and more compatible. Less monkey patching.

I guess I had thought it would be less overhead since these events are already generated by sql and mongo / others. And easier to migrate.

I'll take your input into account and think this through a little more

kbrock:
closing this issue - will use PR to focus on the conversation

PR #211, opened 2016-02-21
Active support notifications

kbrock:
This is what I had in mind for #199 -- leveraging active support notifications.

This is working for me in rails 3 and 5.

`ActiveRecord` already depends upon `ActiveSupport` and is already generating these events. So this doesn't seem to be adding any additional dependencies or overhead.

Left removing the active record patches as a separate commit so can add those back in if you like.

### DIFF
--- a/lib/mini_profiler/profiling_methods.rb
+++ b/lib/mini_profiler/profiling_methods.rb
@@ -8,6 +8,12 @@ module Rack
         c.current_timer.add_sql(query, elapsed_ms, c.page_struct, params, c.skip_backtrace, c.full_backtrace)
       end
 
+      def report_reader_duration(elapsed_ms, row_count = nil, class_name = nil)
+        return unless current && current.current_timer
+        c = current
+        c.current_timer.report_reader_duration(elapsed_ms, row_count, class_name)
+      end
+
       def start_step(name)
         return unless current
         parent_timer          = current.current_timer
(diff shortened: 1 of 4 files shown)


kbrock:
Aah, I understand now, the old mechanism works with pure pg gem (not active record)

Would this be a viable alternative to just the ActiveRecord patch?

SamSaffron:
Yes that is critical, also the AR notifications do not break up the "fetch duration" from the "Execution duration" 

Also I found AR notifications to add a fair amount of overhead.

kbrock:
@SamSaffron Suggestions how I can measure the notification overhead?

If we were adding additional notifications, I could see how that would add some overhead.
But since this is just capturing existing notifications, I'm surprised by this.

To be honest, I thought this would be less overhead than the monkey patch. But I will need to dive deeper on this one.

Any suggestions on how to measure this would be great. The overhead would be so small, not sure if it would be bigger than the variation of a typical rails page (it varies so much)

kbrock:
Changed this to not update the version people are using.

A custom loader (something like #254 ) will allow me to run this and not alter behavior for others

kbrock:
rebased

@SamSaffron This only takes effect if someone explicitly asks for it. I'll try and figure out a way to better dissect the timings.

`Model.distinct.pluck(:name)` does not instantiate, so it has no row count. (Although, not sure if that is the same with pg patch as well.

But this should not affect anyone.
Then again, this is less critical for me with that loader, I can patch in any class I want.

SamSaffron:
definitely I only want AR notifications as a last resort if nothing else works, it is less efficient than our standard monkey patching, I am closing this for now, sorry :(

Issue #196, opened 2016-01-29
Rails 5 `serve_static_assets` deprecation warning


jayfredlund:
I'm running Rails 5.0.0.beta1.1 and using the master branch of this gem (for Rack 2 support) and running into this message:

```Bash
DEPRECATION WARNING: `serve_static_files` is deprecated and will be removed in Rails 5.1.
Please use `public_file_server.enabled` instead.
 (called from serves_static_assets? at /Users/jay/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-mini-profiler-cb7a6cf4d4cd/lib/mini_profiler_rails/railtie.rb:72)
```

The code was committed in https://github.com/MiniProfiler/rack-mini-profiler/commit/ddc193ed77cb3a8db7e44b9a5c404649252fb7af and found in the master branch here: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler_rails/railtie.rb#L68

The deprecation and change is discussed in https://github.com/rails/rails/pull/22173

I'm not sure that adding another `if respond_to?` is a desired solution or what your overall strategy is for supporting multiple versions of Rails going forward, but figured I'd document it here. I can propose something basic, unless someone else is already working on little things like this.

xanview:
Seems this deprecation happens from a few other places also:

```
14:36:15 sidekiq.1 | DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <class:Connection> at /usr/local/lib/ruby/gems/2.3.0/bundler/gems/rack-mini-profiler-a073d1d33a7c/lib/patches/db/mongo.rb:11)
14:36:16 web.1     | DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <class:Connection> at /usr/local/lib/ruby/gems/2.3.0/bundler/gems/rack-mini-profiler-a073d1d33a7c/lib/patches/db/mongo.rb:11)
```

kbrock:
Module prepend was introduced in ruby 2.0.0.

What is the status for supporting 1.9.3 and 1.8.7?

jayfredlund:
[...]

PR #213, opened 2016-03-02
Removes `alias_method_chain` in Mongo patch for Rails 5 deprecation warning

jayfredlund:
Rails 5 issues a deprecation warning when using `alias_method_chain` favoring `Module#prepend`.  `Module#prepend` wasn't introduced until Ruby 2.0, but since we still support 1.9.3, we can't easily switch.

Since `alias_method_chain` was just encapsulating the pattern of using two `alias_method` calls, we can remove the deprecation warning and support Ruby 1.9.3 by just using the two `alias_method` calls that were created.  I've left a `TODO:` in the code to note that we should switch this out to use `Module#prepend` once Ruby 1.9.3 support is dropped.

@kbrock is correct in his comments that using `Module#prepend` and `alias_method_chain` would have to be structured very differently and once I started down that road, it didn't seem to be worth it -- at least at this point in time.  We're not *losing* anything by switching to two `alias_method` calls, though we aren't *gaining* much either (a single deprecation warning).

This addresses the comments brought up in #196 (which can now be closed since the main issue was resolved in #201).

### DIFF
--- a/lib/patches/db/mongo.rb
+++ b/lib/patches/db/mongo.rb
@@ -8,5 +8,9 @@ class Mongo::Server::Connection
     end
     return result
   end
-  alias_method_chain :dispatch, :timing
+
+  # TODO: change to Module#prepend as soon as Ruby 1.9.3 support is dropped
+  alias_method :dispatch_without_timing, :dispatch
+  alias_method :dispatch, :dispatch_with_timing
+
 end


kbrock:
Pretty straight forward. thanks

LGTM

SamSaffron:
sure seems fine to me!

Issue #215, opened 2016-03-09
NoMethodError: private method `singleton_class' called for Rack::MiniProfiler:Class


mockdeep:
We have a stub in our specs just to test that the profiler gets enabled and with the recent upgrade we saw the spec start breaking:

```
     NoMethodError:
       private method `singleton_class' called for Rack::MiniProfiler:Class
````

The spec looks like:

```ruby
expect(Rack::MiniProfiler).to receive(:authorize_request)

get(:index)
```

SamSaffron:
jeez i hate stubs :) care to try a PR to fix it?

seanlinsley:
Perhaps adding `and_call_original` to the stub would fix it. Unless you're trying to disable authentication?

If you are trying to disable authentication, it seems like the better option is to actually disable it as part of your test environment setup.

mockdeep:
We're just trying to verify that it gets called. I don't think adding `and_call_original` would do the trick in this case, since rspec stubs out the method either way. The `singleton_class` method was made private in a recent commit on `rack-mini-profiler`, which is a method rspec uses for stubbing. I'll see if I can get a fix in, unless somebody else wants to tackle it.

seanlinsley:
That being: #190.

@SamSaffron why not drop support for Ruby 1.9? [Ruby 2.0 is no longer supported](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/). For those still on older versions, they wouldn't be upgrading gems for the same reasons they aren't upgrading Ruby.

Or actually, [`singleton_class` is available as early as 1.9.3](http://ruby-doc.org/core-1.9.3/Object.html#method-i-singleton_class) so we don't need to drop support for Ruby versions just for this change.

seanlinsley:
There is [this in the gemspect](https://github.com/MiniProfiler/rack-mini-profiler/blob/77b2c67105278e3eada60da942b8f2aabca36bec/rack-mini-profiler.gemspec#L22) for Ruby 1.8, but Travis isn't actually running 1.8. I assume 1.8 is no longer supported?

SamSaffron:
Yes ... 1.8 is not longer supported

davebrace:
I ran into the same thing after upgrading. Good point about not needing to stub. I updated our specs to just use `Rack::MiniProfiler.request_authorized?` to check after the request if it was authorized or not.

For the example above you could do this: 

```ruby
get(:index)

[...]

PR #217, opened 2016-03-14
Remove singleton_class method

seanlinsley:
It's not needed since Ruby < 1.9.3 isn't supported.

#215

### DIFF
--- a/lib/mini_profiler/profiling_methods.rb
+++ b/lib/mini_profiler/profiling_methods.rb
@@ -110,11 +110,11 @@ module Rack
       end
 
       def profile_singleton_method(klass, method, type = :profile, &blk)
-        profile_method(singleton_class(klass), method, type, &blk)
+        profile_method(klass.singleton_class, method, type, &blk)
       end
 
       def unprofile_singleton_method(klass, method)
-        unprofile_method(singleton_class(klass), method)
+        unprofile_method(klass.singleton_class, method)
       end
 
       # Add a custom timing. These are displayed similar to SQL/query time in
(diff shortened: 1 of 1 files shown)


SamSaffron:
sort of, still would prefer to keep helper around for now. 

```ruby
irb(main):007:0> (class << BasicObject.new; self; end)
=> #<Class:#<BasicObject:0x007fc50b20e810>>
irb(main):008:0> BasicObject.new.singleton_class
NoMethodError: undefined method `singleton_class' for #<BasicObject:0x007fc50b07b250>
	from (irb):8
	from /Users/sam/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'
irb(main):009:0>
```

seanlinsley:
It seems unlikely to me that someone is going to profile a BasicObject. But if you think that's valid and should work automatically, then we have two options:
- make the method public instead of private
- open a bug on RSpec, since it seems like it's using `public_send` when it should be using `send`

SamSaffron:
There is one more option here :) why do all the stubbing and mocking in
that spec? Why not simply test that it correctly profiles a method?

kbrock:
a) I think checking that a particular method is called is a little fragile. Checking the side effects seems better.
b) I'm not sure about the `new` in `BasicObject.new.singleton_class`. Had viewed this method more for setting and detecting class level methods.

My current use case is to have an initializer with a long list of class names and methods to profile:

```ruby
[[class1, :method1],[class2, :method2]].each do |klass, method|
  if klass.respond_to?(method)
    ::Rack::MiniProfiler.profile_singleton_method(klass, method) { |a| "#{klass.name}.#{method}" }
  else
    ::Rack::MiniProfiler.profile_method(klass, method) { |a| "#{klass.name}##{method}" }
  end
end
end
```

SamSaffron:
I'm closing this, I don't think we need to make any changes here just to support people testing MiniProfiler internal implementation in their spec suite ... its just fragile and not supported. 

I did not pick up that this was someone else's spec suite and not our, apologize for that.

PR #218, opened 2016-03-14
Ruby updates

seanlinsley:
Written as part of #217 for #215

#### Don't include Gemfile.lock in source control

It's common to exclude the lockfile from source control because it lets CI test against new gem releases as they come out.

Along with this change I've version-constrained RSpec to 2.14. I didn't move that to 2.99 because of the deprecation warnings it adds.

####  Specify Ruby 1.9.3 as the minimum supported Ruby version
and remove some Ruby 1.8 support code


SamSaffron:
sure, thanks!

PR #221, opened 2016-03-18
small documentation change for flamegraph_sample_rate configuration

dnasseri:
Issue here: https://github.com/MiniProfiler/rack-mini-profiler/issues/222
The configuration documentation for flamegraph_sample_rate is a bit confusing, and makes it look like it accepts a string in the form of `'0.1ms'` as opposed to a float `0.1`.


nateberkopec:
Maybe change the description too: "How often to capture stack traces for flamegraphs _in milliseconds_" or something like that.

dnasseri:
@nateberkopec updated the description as per your description

SamSaffron:
Sure, looks good!

PR #225, opened 2016-03-24
Just using defined

kbrock:
I would imaging that just using `defined?` would be sufficient to know if a gem was imported without
resorting to testing whether it is a `class` or `module`.

Well, that is my interpretation of the code that introduced this method: af73d1255473af0cff9c5f02b429a49c755a0597

But it is very possible that I don't understand the nuances that you were targeting.
Thanks

### DIFF
--- a/lib/patches/sql_patches.rb
+++ b/lib/patches/sql_patches.rb
@@ -8,20 +8,8 @@ class SqlPatches
     @patched = val
   end
 
-  def self.class_exists?(name)
-    eval(name + ".class").to_s.eql?('Class')
-  rescue NameError
-    false
-  end
-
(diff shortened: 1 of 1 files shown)


SamSaffron:
Probably, but this is a very big change I worry that we are missing an edge case or 2

kbrock:
fair enough - but the original code looks like it called this instead of `defined?(ActiveRecord)`
Maybe the original code would have worked with this, but possibly the newer code would not.

Just close it then?

SamSaffron:
yeah going to close this ... worried it is too risky

kbrock:
thanks

PR #226, opened 2016-03-30
Attempt to address discourse in a directory issue #25

nspring:
Sam: This is a suggested patch to try to handle the Discourse issue that revived #25.  It does not break my install (the whole application is run in a directory by passenger), and assumes both that the prior code worked and was necessary (action_controller.instance was defined for the mounted engine) so that all that is necessary to fix the issue is to compose the two variables.

How to run discourse in this configuration seemed pretty involved and under-documented, so reproducing the bug seemed tricky with nothing but your comment on #25 to go on.


SamSaffron:
Thanks, let's try it out, looks like it could work

cc @foobarwidget

Issue #139, opened 2015-05-18
Parse meaningful data from stored results


YoranBrondsema:
Hi,

I have an Ember.js application backed by a Rails API. Since the `index.html` is generated by the front-end, MiniProfiler cannot inject itself onto the page. However, I would still like to use its results.

It seems the output in `/tmp/miniprofiler/*` cannot be used without parsing it. Is there a way to store the same output that it prints in HTML in a file? Then I can just cURL to actions in my controller and view a meaningful MiniProfiler report.

silasb:
:+1: I am in a similar situation where it would be helpful to get profiler information (and flamegraphs) for an mobile API that I'm working on.

Aldekein:
+1, I'd like to profile backend for an angular application. All the interaction between frontend and rails is in JSON format and I'd like to parse the mp_timers_* files.

Aldekein:
@silasb @YoranBrondsema 
I get something that could be human-read if I run the command in rails console and then replace \n with new-lines in Sublime and set View->Syntax->YAML:

```
YAML::dump(Marshal.load(::File.open('/var/www/project.lc/tmp/miniprofiler/mp_timers_z4h2y3sy5f662liqu2n5',"rb") {|f| f.read}))
```

kevinjalbert:
I would love to see some way to get more details on API requests that we feed into a Javascript front-end (i.e., Ember/Angular).

I've seen https://github.com/dejan/rails_panel, which does something like this via a chrome extension, which is an interesting approach.

SamSaffron:
You should be able to inject it just fine:

at the moment this is the extra script that is added: 

```
 <script async type="text/javascript" id="mini-profiler" src="/mini-profiler-resources/includes.js?v=12b4b45a3c42e6e15503d7a03810ff33" data-version="12b4b45a3c42e6e15503d7a03810ff33" data-path="/mini-profiler-resources/" data-current-id="redo66j4g1077kto8uh3" data-ids="redo66j4g1077kto8uh3" data-position="left" data-trivial="false" data-children="false" data-max-traces="10" data-controls="false" data-authorized="true" data-toggle-shortcut="Alt+P" data-start-hidden="false" data-collapse-results="true"></script>
```

Each time we change js versions change... but overall once you include this ... all new requests to API will be added. The version stuff is there to break caching you could simply put something random there every time you build.

kevinjalbert:
[...]

PR #227, opened 2016-04-05
Update README on how to enable speed badge in SPA

kevinjalbert:
Addresses comment to beef up README instruction on getting MiniProfiler working on SPAs -- #139

-----

An SPA needs a little more attention to get MiniProfiler's speed badge to appear. This commit adds more information on how to inject the necessary script to load the speed badge.

The referenced GitHub issue's comment is also included in the new information for context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,12 @@ window.MiniProfiler.pageTransition();
 
 This method will remove profiling information that was related to previous page and clear aggregate statistics.
 
+You also need to inject the following in your SPA's `index.html` to load MiniProfiler's speed badge ([extra details surrounding this script](https://github.com/MiniProfiler/rack-mini-profiler/issues/139#issuecomment-192880706)):
+
+```html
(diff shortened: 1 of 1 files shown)


SamSaffron:
Need to be a little bit less strong with wording there, if SPA is served not via the Rails stack this is required (for Discourse for example, it simply works out of the box and we are an SPA)

kevinjalbert:
I'm afraid I don't quite understand. Is it the mention of adding the script to the `index.html`? I guess it doesn't necessarily have to be on that, as long as the script is loaded somehow within the SPA.

Or am I missing something here? In my particular use of this, I have a Rails backed API along side a separate ember SPA. I had to add the mentioned script to ember's generated `index.html` so that the speed badge would appear, is it possible to somehow have this work _out of the box_?

SamSaffron:
Discourse does not use ember-cli, we server all our pages via Rails for the SPA.

I think this is specific to "Including rack-mini-profiler on pages that are not generated via Rails"  maybe a sub heading for that.

kevinjalbert:
Oh! I see. I mis-read your earlier response and got confused on whether it was served by rails or not. I'll add a sub heading as suggested.

kevinjalbert:
@SamSaffron I've updated the wording to reflect your suggestions.

SamSaffron:
perfect, thanks... the guid though will change with every release of MP maybe put a note about that (it will still work but you need to change the guid every time you want to expired the script)

kevinjalbert:
@SamSaffron good point. Just updated.

SamSaffron:
cool! thanks for this!

PR #228, opened 2016-04-08
Fix PATH_INFO starting with double slashes

### LINKED ISSUE #114: Mounted Routes

haggen:
Apparently when using the profiler from mounted routes the variable `env['PATH_INFO']` would come starting with two slashes `//` which caused conditionals like `starts_with?` and other comparisons to fail.

Tests ran OK except the ones that needed redis and memcache - since I didn't have those installed.

Fixed #114.

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -120,7 +120,8 @@ module Rack
     end
 
     def serve_html(env)
-      file_name = env['PATH_INFO'][(@config.base_url_path.length)..1000]
+      path      = env['PATH_INFO'].sub('//', '/')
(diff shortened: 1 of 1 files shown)


SamSaffron:
any chance we can add some specs for this, it always breaks.

haggen:
Should I just test a `PATH_INFO` starting with double slashes?

SamSaffron:
ok lets try it then

PR #231, opened 2016-04-20
Module exists

kbrock:
Don't know what it is about this code, but it always distracts me.
Here is another attempt to try and reduce it down.

What do you think?

Thanks

### DIFF
--- a/lib/patches/sql_patches.rb
+++ b/lib/patches/sql_patches.rb
@@ -1,5 +1,9 @@
 class SqlPatches
 
+  def self.unpatched?
+    !patched?
+  end
+
   def self.patched?
     @patched
   end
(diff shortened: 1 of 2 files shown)


SamSaffron:
originally I did this for perf, to avoid all the unconditional evals whats the bench diff here.

kbrock:
class defined|pure eval|`define?`/`eval`
-----|---|---
yes|140k/s|180k/s
no|150k/s|6M/s
 
Fun deep dive [gist](https://gist.github.com/kbrock/edd05b8e4b463eda07082e2efa33e6f2) - When I ran the tests via `time bundle exec rake` I did not see any difference. It is only called 12 times after all.

I prefer minimizing the reading in the line, so I like just plain `defined?`, but if performance really is a concern for you here, I could put in a PR for a much faster version - e.g.:

```ruby
def module?(k)
  k.class == Module
end

defined?(Benchmark) && module?(Benchmark)
```

SamSaffron:
k closing this one

PR #233, opened 2016-04-21
Define directly

kbrock:
You said performance...

Here is an alternative to #231 that you may like:

- Much more performant
- Reduces source code
- Removes the `eval` all together
- levarages `unpatched` from #232.

### Algorithm

Instead of having a method to check the class, just define it inline. And don't eval it.

### Benchmark

class defined|before|after
-----|---|---
yes|180k/s|6.328M/s
no|6M/s|7.565M/s

Details can be found in [benchmark]. Before is using `define?` and `eval`. After is `defined?` and a direct `.class` call.
 
[benchmark]: https://gist.github.com/kbrock/edd05b8e4b463eda07082e2efa33e6f2

### ASIDE

This is getting similar to #225 (which I still like most of all the choices)

If another of the ideas from the [benchmark]() tickle your fancy, let me know.

[benchmark]: https://gist.github.com/kbrock/edd05b8e4b463eda07082e2efa33e6f2

### DIFF
--- a/lib/patches/sql_patches.rb
+++ b/lib/patches/sql_patches.rb
@@ -1,4 +1,7 @@
 class SqlPatches
+  def self.unpatched?
+    !patched?
+  end
 
   def self.patched?
     @patched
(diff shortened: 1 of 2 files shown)


SamSaffron:
sure lets go with this!

SamSaffron:
less evals are better totally happy with that, we could possibly extract a method here, but will leave it like this for now.

PR #235, opened 2016-04-21
Grouped the view params together and added setters

kbrock:
I tend to like children expanded by default.

This adds the remaining config params for the ui's properties

thanks


SamSaffron:
sure, we also should update the readme with the new settings.

PR #236, opened 2016-04-21
Organize controller

kbrock:
I'd like to understand the main controller loop better (i.e.: `def call()`)

This PR is the start of extracting a command / factory pattern. A lot of places in the code are very tight. This sticks with the functional approach, focusing on defining the boundaries and parameters of the various commands.

Does this look like a good course of action or a bit too disruptive?
Any special hints before I go too far?

Thanks

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -151,6 +151,18 @@ module Rack
       ClientSettings.new(env)
     end
 
+    # @return :skip_it if this requst should be skipped
+    def determine_action(client_settings, env)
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am technically ok with this work, I just am very very very afraid of regressions, can you ensure there is full coverage for everything you are changing?

seanlinsley:
This needs to be rebased

kbrock:
I'm intimidated by adding tests here.

SamSaffron:
closing this for now as too scared to change it

PR #237, opened 2016-04-22
access environment from passed values

kbrock:
This is blind, and more of a question than anything else.

@nspring Is there a reason for accessing `ENV` instead of `env` here?
This is the only use of `ENV` in the code base and potentially makes the tests fragile to modify the global environment.

Thanks for any insight


nspring:
Honestly, I started with little env; when running under apache / passenger, env[] didn't have the value.  Big ENV did.  If the problem is that running the test is potentially destructive to the environment for later tests, could we restore the prior value at the end of the test (ensure) instead?

kbrock:
thanks @nspring - that is what I wanted to know. hasn't caused a problem up until now

PR #238, opened 2016-04-24
fix multi edit fail for Neo4j

kbrock:
So sorry this slipped through. (sublime multi cursor fail)

Do you have ideas on how to prevent this error?
Define all of these patches in a module and include into classes if they are defined? (instead of direct monkey patches?)

e.g. make more similar to activerecord.

Is there a point in time that you want to drop 1.9.3 and use module prepend? (sorry, can't remember your stance here)


SamSaffron:
oops

regarding 1.9.3 support going to keep it for a bit longer... maybe at end of the year we can revisit ... 

we could make test cases for this but would have to be very careful not to infect global state

PR #244, opened 2016-05-17
active record add instrumentation in common way

kbrock:
a few thoughts:

1. I like the way this adds a module rather than monkey patching the classes directly.
2. This can be easily changed to module prepend
3. The pattern seems to be to check `patched?` and defined? outside the class.
4. Not sure why we are ensuring `Rails` is defined.

Questions:

1. Would it make sense to have all the patches work the same way?
(I'd be happy to implement this, but would be flying by the seat of my pants as I do not have most of these systems installed)
2. At a future time, would we want to remove the inject logic from the bottom of the patch file. Instead, all patches could be required, but sql_patches would apply the desired patch.

### DIFF
--- a/lib/patches/db/activerecord.rb
+++ b/lib/patches/db/activerecord.rb
@@ -39,15 +39,13 @@ module Rack
         rval
       end
     end
-  end
 
-  def self.insert_instrumentation
-    ActiveRecord::ConnectionAdapters::AbstractAdapter.module_eval do
-      include ::Rack::MiniProfiler::ActiveRecordInstrumentation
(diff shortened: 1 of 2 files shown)


kbrock:
>At a future time, would we want to remove the inject logic from the bottom of the patch file. Instead, all patches could be required, but sql_patches would apply the desired patch.

The main driver on this is to add test coverage. Interesting, didn't know you had a bot doing that. fun

kbrock:
If the main application supports ActiveRecord and it does not need Rails, then this patch should work fine.

Another PR will drop this requirement

PR #245, opened 2016-05-18
Add host name and process id to server time output

jhicks:


SamSaffron:
I am fine with this change but it is going to have to be opt in, adds a lot of noise to the UI in many cases.

seanlinsley:
@jhicks ping

kbrock:
@seanlinsley I'm confused with the ping.
This looks like a nice feature, but it adds too much noise for the average user.
So the request is: could you change this to add an option and make it opt in?

kbrock:
Think test failure is bogus. Should be fixed by #255

kbrock:
Also, could you include a before and after screen shot?
Thanks

SamSaffron:
I am closing this, for me to accept this it is going to have to be an opt in thing :( sorry.

PR #249, opened 2016-06-19
Fix Travis builds failing due to Guard dependencies

dgynn:
This PR gets Travis tests working for all Rubies by moving the `guard` development dependency to the Gemfile so `:platforms` can restrict it to Ruby 2.2+.

It also adds Ruby 2.3 to the build matrix and ensures `bundler` is up-to-date enough to know about `:mri_22` and `:mri_23` platforms.


SamSaffron:
excellent

Issue #70, opened 2014-03-24
Not showing in Rails 4


abuisman:
When I include the gem in my rails app the profiler is not shown. However, the javascript is included and the div is created for the widget, but the final content is empty.

I tried adding `Rack::MiniProfiler.authorize_request` in a before_filter in my application controller: no dice.

I tried creating an initializer and add `require: false` in the gem file as described in the README: no profiler.

I am using puma, rails 4.0.1, and ruby ruby 2.0.0p247

Any tips?

akshetpandey:
+1, removing oj_mimic_json and following what @hakunin said solved the issue.

pencilcheck:
+1, removing oj_mimic_json works. It sucks because it was suppose to make json serialization faster.

SamSaffron:
this is all very strange can someone write a test case that fails here, somehow ojs json implementation is lacking and its hurting us here.

akshetpandey:
The issue is that OJ doesn't handle circular reference in objects. For example:

```ruby
2.1.5 :001 > require 'oj'
 => true
2.1.5 :002 > require 'oj_mimic_json'
 => true
2.1.5 :082 > b = {a: {}}
 => {:a=>{}}
2.1.5 :083 > b[:a] = b
 => {:a=>{...}}
2.1.5 :084 > b.to_json
NoMemoryError: Too deeply nested.

	from (irb):84:in `to_json'
	from (irb):84
	from /Users/akshet/.rvm/rubies/ruby-2.1.5/bin/irb:11:in `<main>'
```

pencilcheck:
Oj.default_options = {
  :mode => :compat,
  :use_to_json => true
}

This works for me for circular references

SamSaffron:
before any of this can someone help isolate the circular ref, so we can remove it, we don't really need it in our persistence structure.

DavidRagone:
I also had to both remove Deflator and oj_mimic_json to get this to work. Ruby 2.2, Rails 4.2.1

eggie5:
+1 for removing `oj_mimic_json` gem.

kevinjalbert:
Just wanted to chime in that I encountered the same issue. I had to remove the `oj_mimic_json` gem for this to work.

The solution that @pencilcheck suggested didn't work for me.

jhubert:
I didn't have the `oj_mimic_json` gem, but removing `heroku-deflator` worked for me.

thatpixguy:
[...]

PR #250, opened 2016-06-19
Resolve circular reference issue while generating JSON with Oj

dgynn:
This PR addresses the problem identified in #70 where `oj_mimic_json` prevents the JSON response from being created.

Oj prefers `as_json` to `to_json` and previously the TimerStruct classes only implemented `to_json`. This commit adds `as_json` support that matches the `to_json` implementation so either way will work.

Specifically, the circular reference is between Page and Request (TimerStruct) objects where Page has reference to a `@attributes[:root]` Request and Request has a `@page` attribute.

I have not added test cases for this new code because `as_json` will only work if ActiveSupport extensions have loaded so that Hash has `as_json` support. The fix can be verified by adding the `oj_mimic_json` gem to an existing app.

### DIFF
--- a/lib/mini_profiler/timer_struct/base.rb
+++ b/lib/mini_profiler/timer_struct/base.rb
@@ -27,6 +27,9 @@ module Rack
           ::JSON.generate( @attributes, :max_nesting => 100 )
         end
 
+        def as_json(options = nil)
+          @attributes.as_json(options)
+        end
       end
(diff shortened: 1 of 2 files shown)


SamSaffron:
cool!

kbrock:
Thanks so much. Many people wanted to get OJ working

PR #251, opened 2016-06-23
use File.exist? instead of deprecated File.exists?

y-yagi:
`File.exists?` was deprecated in Ruby 2.1. Ref: https://github.com/ruby/ruby/blob/v2_1_2/file.c#L1413


kbrock:
:+1: `exists?` has been marked obsolete since 1.8.6 and possibly before [link].
Lets remove it.

[link]: http://ruby-doc.org/core-1.8.6/File.html#method-c-exist-3F

SamSaffron:
sure, thanks, sorry for the delay

y-yagi:
Thanks!

PR #253, opened 2016-07-01
pull out full_stacktrace logging logic

kbrock:
Noticed `full_backtrace` was on each of the lines for trimming the backtrace.
So I split that logic to what I find easier to read.

In the end, this this is just a minor change.
(I had bigger plans like using `Regexp.join()` didn't get the performance boost I had wanted.)

Thanks


kbrock:
failure fixed by #255

SamSaffron:
I am closing this for now, the specs seem to be failing and it is cosmetic only

PR #254, opened 2016-07-01
Allow environment variable to define sql patching

kbrock:
When users introduce their own patches, or want to work around
patch detection, this allows them to override with
`$RACK_MINI_PROFILER_PATCH`

setting to `"false"` skips sql patches all together

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,29 @@ class MyApp < Sinatra::Base
 end
 ```
 
+#### Patching ActiveRecord
+
+A typical web application spends a lot of time querying the database. rack_mini_profiler will detect the ORM that is available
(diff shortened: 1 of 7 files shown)


kbrock:
failure fixed by #255

SamSaffron:
I think this is fine, can you document this in the readme and if possible add a test case for it?

kbrock:
1. added `README`
2. There are a number of test cases. any other one of interest
3. Modified `patch` to allow an initializer to pass particular ORM names.

SamSaffron:
awesome, also love the doco

kbrock:
thanks so much

PR #255, opened 2016-07-01
For older version of ruby, use older rack

kbrock:
Rack `2.0.1` requires a more recent version of ruby.

Trying to work around this and still support  older versions of ruby


SamSaffron:
happy to try this out, 2.0 is on it's dying legs

PR #256, opened 2016-07-04
Fix two dates in Chanelog

agrimm:
A fairly cosmetic change, but it makes reading the Changelog less confusing.


agrimm:
Another change that's somewhat cosmetic, but makes it less confusing: making sure all versions include the major number (0).

SamSaffron:
Sure, thanks!

PR #257, opened 2016-07-04
Use earlier version of json for ruby 1.9.3

kbrock:
Current version of the json gem requires ruby 2.0


SamSaffron:
I kind of want to drop 1.9, but we can do this for now

kbrock:
Hey. Dropping 1.9 sounds great. I want things to pass. And trying not to push my own agenda (too hard) ;)

Issue #242, opened 2016-05-11
Rails5 and Deflate middleware


dreyks:
So I've been migrating an app to Rails5 and rack mini profiler just stopped showing up.
The 

``` ruby
require 'rack-mini-profiler'

Rack::MiniProfilerRails.initialize!(Rails.application)

Rails.application.middleware.delete(Rack::MiniProfiler)
Rails.application.middleware.insert_after(Rack::Deflater, Rack::MiniProfiler)
```

does not work anymore, due to [this commit](https://github.com/rails/rails/commit/a39498ae060f07132ddbbc45933a819686f0eb24) in rails. Now the `delete` operations are issued at the end

dreyks:
Ok, this, though being ugly, works

```ruby
Rails.application.middleware.swap(Rack::Deflater, Rack::MiniProfiler)
Rails.application.middleware.swap(Rack::MiniProfiler, Rack::Deflater)
```

seanlinsley:
@dreyks could you open a PR to update the readme?

dreyks:
#258 seems to fix this. I'll test it and close the issue

ChallahuAkbar:
So, was this fixed then?

OsamaSayegh:
Yes this was fixed thanks to https://github.com/MiniProfiler/rack-mini-profiler/pull/258. I'm closing this issue but feel free to open a new one if you still see this. Thanks!

PR #258, opened 2016-07-11
Disable compression on profiled requests for simpler configuration

dgynn:
The middleware ordering when `Rack::Deflater` is in use still seems to be causing issues for a lot of people #70. And Rails 5 introduces a new difficulty in changing the stack order #242.

This PR works around the problem by simply disabling compression on profiled requests. This is a similar approach to disabling caching. Only profiled requests would have compression disabled.

With this approach, all of the configuration hoops around re-ordering the middleware could be removed and it should (hopefully) enable more people to use the automatic initialization. It would also be possible to add another configuration option like `config.disable_compression` just like the existing `config.disable_caching`.

Let me know what you think of this approach and/or if it should be a config setting. If it is a go, I can update the readme and changelog.

BTW, it would be possible in the railtie to use an initializer that runs *after* the stack has been created to decide whether to disable compression.
```ruby
initializer "rack_mini_profiler.after_build_middleware", :after => :build_middleware_stack do |app|
  middlewares = app.middleware.middlewares
  if middlewares.include?(Rack::Deflater) && middlewares.index(Rack::Deflater) > middlewares.index(Rack::MiniProfiler)
    Rack::MiniProfiler.config.disable_compression = true
  end
end
```

### DIFF
--- a/lib/mini_profiler/config.rb
+++ b/lib/mini_profiler/config.rb
@@ -18,7 +18,7 @@ module Rack
       :flamegraph_sample_rate, :logger, :pre_authorize_cb, :skip_paths,
       :skip_schema_queries, :storage, :storage_failure, :storage_instance,
       :storage_options, :user_provider
-    attr_accessor :skip_sql_param_names, :max_sql_param_length
+    attr_accessor :skip_sql_param_names, :suppress_encoding, :max_sql_param_length
 
     # ui accessors
     attr_accessor :collapse_results, :max_traces_to_show, :position,
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -253,6 +253,10 @@ module Rack
           env['HTTP_IF_NONE_MATCH']     = ''
         end
 
+        orig_accept_encoding = env['HTTP_ACCEPT_ENCODING']
+        # Prevent response body from being compressed
+        env['HTTP_ACCEPT_ENCODING'] = 'identity' if config.suppress_encoding
+
         if query_string =~ /pp=flamegraph/
           unless defined?(Flamegraph) && Flamegraph.respond_to?(:generate)
 
@@ -279,6 +283,7 @@ module Rack
         end
       ensure
         trace.disable if trace
+        env['HTTP_ACCEPT_ENCODING'] = orig_accept_encoding if config.suppress_encoding
       end
 
       skip_it = current.discard
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -88,6 +88,17 @@ module Rack::MiniProfilerRails
       Rack::MiniProfilerRails.initialize!(app)
     end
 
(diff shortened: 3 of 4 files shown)


SamSaffron:
I am a bit uneasy about having this impact compression on people doing compression in NGINX. Would much prefer if the railstie did this conditionally based on Deflater being in the stack after mini profiler

dgynn:
Yeah, I'm not thrilled with disabling compression either and I do want to make sure this doesn't affect non-profiled requests.

I checked on the NGINX compression and this didn't affect it (using `gzip on` and `proxy-pass`). The `Accept-Encoding` parameter is an inbound request parameter. NGINX would still know that the client had passed something like `Accept-Encoding: gzip, deflate, sdch` in the headers. Only in the Rack `env` would the header value have been overwritten. I haven't checked to see that the same is true with Passenger. But anything that proxies back to a Rack server should be OK.

There is an outbound response header for `Cache-Control: no-transform` that tells upstream servers not to modify the response, including doing gzip encoding. Since this doesn't modify any response headers I believe everything should be fine.

All that being said, I can still do it using the Railtie initializer to check explicitly for Rack::Deflater lower in the stack. Or... another simple change would be to restore `env['HTTP_ACCEPT_ENCODING']` to the original value after the call returns to the profiler and before it hands it back up the stack.

I'll look into both of those and see about updating the PR this week.

dgynn:
@SamSaffron I've updated the PR to do middleware stack-ordering detection in an initializer and to restore the `env['HTTP_ACCEPT_ENCODING']` value. 

The new config setting is `suppress_encoding` and is primarily meant to be for internal use. It wouldn't need to be set by users although it could if some other middleware (not Rack::Deflater) was causing a similar issue. In cases where the railtie doesn't run, the config setting defaults to nil so the current behavior is maintained.

Thoughts? I'm OK closing this if you think it could create more problems than it solves.

SamSaffron:
I think this change is fine, going to merge it in. 

Thanks!

dreyks:
this breaks other gems that manipulate middleware stack from initializers. the error is `can't modify frozen Array (RuntimeError)`

I think there's something wrong with rails railtie itself rather than miniprofiler. It seems that `initializer` without `after: build_middleware_stack` are still executed after the miniprofiler initializer, which calling `app.middleware.middlewares` freezes the middleware stack

dgynn:
@dreyks Can you give an example of a gem that is causing the problem? Also, which version of Rails did you see this with?

dreyks:
[request_store](https://github.com/steveklabnik/request_store/blob/master/lib/request_store/railtie.rb#L5)
Rails 5

dreyks:
also web_console. Really any gem that has its own middleware fails. The immediate workaround is to move `gem 'rack-mini-profiler'` to the bottom of Gemfile

dgynn:
@dreyks Thanks for finding this. I've just pushed up #261 to address the issue. The new initializer with the `:after` was causing `:build_middleware_stack` to execute earlier. It does seem like the Rails "Finisher" initializers should always be at the end of the initializer list but that isn't the case.

PR #261, opened 2016-07-25
Move middleware ordering detection to after_initialize block

dgynn:
This resolves the issue with #258 that is causing `:build_middleware_stack` to be executed before all middleware has been configured.

Using an initializer with `:after => :build_middleware_stack` causes the :build_middleware_stack initializer to be pulled forward in the initializer ordering rather than pushing the rack-mini-profiler initializer to execute later. Doing this in as an `after_initialize` ensures it happens later in the process.


SamSaffron:
sure, seems good!

PR #264, opened 2016-09-06
Permit jQuery3

existent-co-uk:


SamSaffron:
Sure, we should allow 3, have you tested it?

PR #266, opened 2016-09-15
README Use the more idiomatic `#development?`

fny:
Minor nitpick. I won't be offended if you reject. 😄


SamSaffron:
I dunno, I am not a fan of this magic really even though it is more idiomatic.

PR #268, opened 2016-10-06
fixes unstable dependency on ::Rails::VERSION

### LINKED ISSUE #267: Using Sinatra w/ ActiveRecord Breaks `require 'rack-mini-profiler'`
Since `::Rails` exists, but `::Rails::VERSION` does not, `require 'rack-mini-profiler'` breaks.

The error occurs in line 34 of rack-mini-profiler.rb: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/rack-mini-profiler.rb#L34

    uninitialized constant Rails::VERSION (NameError)
    /.rvm/gems/ruby-2.2.1/gems/rack-mini-profiler-0.10.1/lib/rack-mini-profiler.rb:34:in `<top (required)>'

sfsekaran:
Fixes #267

### DIFF
--- a/lib/rack-mini-profiler.rb
+++ b/lib/rack-mini-profiler.rb
@@ -31,6 +31,6 @@ require 'mini_profiler/profiler'
 require 'patches/sql_patches'
 require 'patches/net_patches'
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
nice!

PR #269, opened 2016-10-06
Improve README for profile-memory options and flamegraph dependencies

### LINKED ISSUE #246: Extending Profile with Memory
Hi Sam, what do you think about inserting https://github.com/SamSaffron/memory_profiler as an option like `pp=memoryprofile`. This would be consistent with flamegraphs and would show the output of the memory profiling tool.

I'd be happy to PR this if you would consider it.

dgynn:
This PR primarily adds the docs for using `memory_profiler` with `?pp=profile-memory' query string option.

I also tried to address the changes to the `flamegraph` dependencies that are required now that `fast_stack` is no longer a hard dependency (SamSaffron/flamegraph#11). The combinations of Ruby versions and platforms are a bit complicated to call out succinctly but this should be enough to get Ruby MRI users pointed in the right direction.

Resolves #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ Middleware that displays speed badge for every html page. Designed to work both 
 
 #### Features
 
-* database profiling. Currently supports Mysql2, Postgres, Oracle (oracle_enhanced ~> 1.5.0) and Mongoid3 (with fallback support to ActiveRecord)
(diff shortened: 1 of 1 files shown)


SamSaffron:
awesome, long term what we really want is optional dependency support in rubygems, this is a great example of the problem.

PR #275, opened 2016-11-10
Updates to new turbolinks event name `turbolinks:load`

cannikin:
Perhaps this should bind to both page:change and turbolinks:load to support backwards compatibility?

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -404,7 +404,7 @@ var MiniProfiler = (function () {
         });
 
         if (typeof Turbolinks !== 'undefined' && Turbolinks.supported) {
-            $(document).bind('turbolinks:load.mini-profiler', function() {
+            $(document).bind('page:change.mini-profiler turbolinks:load.mini-profiler', function() {
                unbindDocumentEvents();
             });
         }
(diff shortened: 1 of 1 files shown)


SamSaffron:
does this break backwards compat? Is it a safe change?

eileencodes:
It's not backwards compatible as the new Turbolinks 5 is completely rewritten.

Sam and Javan wrote an FAQ about Turbolinks 5 vs Classic [here](https://github.com/turbolinks/turbolinks/wiki/Turbolinks-5-FAQ#should-i-use-turbolinks-5-or-turbolinks-classic) which provides this [shim](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/compatibility.coffee) if you're interested in compatibility.

cannikin:
Correct, as currently written all I did was replace `page:change` with `turbolinks:load`. That's the only `page:` related event present in the JS. That shim looks like it patches all of turbolinks 5 to fire the same events as turbolinks 3.

rack-mini-profiler seems to want to work out-of-the-box with Turbolinks so I feel like it would be reasonable to default it to "just working" without having to manually add a downgrade compatibility layer. I'll update the code to listen for both events, but you guys let me know if it's something you'd rather leave to the user to handle. If so we should at least update the README to let folks know how to do it.

cannikin:
Any updates on this one?

SamSaffron:
I will merge, thanks @cannikin

PR #278, opened 2016-12-12
Adding option to specify HTML element that UI is injected into

edave:
Added a configuration option in to allow users to have the rack-mini-profiler put somewhere in the DOM besides `<body>` (but it defaults to `<body>` if nothing is provided, so it will behave as expected with existing users). This allows a user to further customize the display/usage of the mini profiler UI in the event that having it at the root level of the DOM makes it inaccessible due to the individual peculiarities of their own app's organization/CSS layout.


SamSaffron:
looks good to me, weird travis issue .. hmmm

edave:
@SamSaffron is there anything we can do on our end? I was having this problem w/ the testing before I even started working on the project, so I don't think it's anything in the PR itself. Thanks

SamSaffron:
I am going to merge it in and see what we can do about getting travis less crazy

PR #279, opened 2016-12-13
c.skip_schema_queries set twice

halfling-rram:
we already have assignment:
c.skip_schema_queries = (Rails.env.development? || Rails.env.test?)


SamSaffron:
sure seems fine, thanks

PR #289, opened 2017-04-01
Avoid some unwanted warnings

chikamichi:
Hi,

While reviewing a verbose version of some RSpec logs for a Rails app, I noticed mini profiler generates a lot of warnings that can be avoided pretty simply.


SamSaffron:
sure, seems fine to me! thanks

Issue #288, opened 2017-03-26
Log bindings for pg prepared statements


antonbaron:
Hi everyone, thx for great gem! 
Anyone knows how to show prepared sql bindings in profiler results?
E.g. show 
`SELECT  "foo".* FROM "foo" WHERE "foo"."bar" = $1 LIMIT $2  [["bar", "baz"], ["LIMIT", 1]]`
instead of just
`SELECT  "foo".* FROM "foo" WHERE "foo"."bar" = $1 LIMIT $2`

Imho it's very important information to show in profiler

SamSaffron:
there is an element of risk there cause the params can be enormous in some cases, I am totally open for a PR that adds this support optionally.

antonbaron:
Ok, we could add a config key "maximum prepared statement param length" to deal with it. 
As far as I understand we can use something like https://github.com/MiniProfiler/rack-mini-profiler/issues/171 . I'll git into this, thx for a reply.

kbrock:
We have [sql_max_param_length](https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler/config.rb#L21)

Thought that was used, is passed into the back end, and has specs too.
The only missing part is the modifications to serialization to the front end and the javascript to render it.

(I use a custom ruby back end script, and am viewing binding variables just find)

SamSaffron:
Nice, totally support fixing this!

SamSaffron:
completed and merged now

PR #291, opened 2017-04-19
Adding sql bindings for pg in profiler results

Neznauy:
Hi,
according to https://github.com/MiniProfiler/rack-mini-profiler/issues/288 , I added frontend to show it. Also, there was no transfer bindings, and bindings had only value and no name, in pg at least.

### DIFF
--- a/lib/patches/db/pg.rb
+++ b/lib/patches/db/pg.rb
@@ -52,7 +52,7 @@ class PG::Connection
     start        = Time.now
     result       = exec_without_profiling(*args,&blk)
     elapsed_time = SqlPatches.elapsed_time(start)
-    record       = ::Rack::MiniProfiler.record_sql(args[0], elapsed_time)
+    record       = ::Rack::MiniProfiler.record_sql(args[0], elapsed_time, get_binds(*args))
     result.instance_variable_set("@miniprofiler_sql_id", record) if result
 
     result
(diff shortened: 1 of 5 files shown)


SamSaffron:
can I see some screenshots of the change?

Neznauy:
max_sql_param_length = 1
eq1
max_sql_param_length = 0, the same as before my changings
eq0
max_sql_param_length = 5
eq5
max_sql_param_length = nil
eqnil

SamSaffron:
looks good! we should probably default param length to 20 or so.

lib/patches/db/pg.rb
```diff
@@ -92,11 +92,29 @@ def async_exec(*args,&blk)
...
-    record       = ::Rack::MiniProfiler.record_sql(args[0], elapsed_time)
+    record       = ::Rack::MiniProfiler.record_sql(args[0], elapsed_time, get_binds(*args))
     result.instance_variable_set("@miniprofiler_sql_id", record) if result
 
     result
   end
-
+  
+  def get_binds(*args)
+    return if args[1].nil?
```
SamSaffron: hmmm this concerns me performance wise, we are building hh without shortening it at all. 

this patch should only grab the text the ui needs

SamSaffron: also lots of regexes here which should be constants, I would also do a while loop here cause perf is super sensitive for profilers

SamSaffron:
Sorry, I am reverting this for now and doing another release, it is somewhat buggy and completely untested, totally happy to add this if we have a proper set of tests for it.

PR #292, opened 2017-05-11
Ensures that `RedisStore#get_unviewed_ids` returns only ids for existing records

sayhiben:
When using rack-mini-profiler configured with `RedisStore`, I began encountering 404 responses when `includes.js` made POST requests to `/mini-profiler-resources/results` after 24 hours of use. These 404 errors were due to the store being unable to load records matching the ids sent by the client because they had expired. The client receives these ids via the `X-MiniProfiler-Ids` header, which is provided by `RedisStore#get_unviewed_ids`.

`RedisStore#get_unviewed_ids` had the possibility of containing references to redis records that were expired. This is because the list of unviewed ids and the individual `PageStruct` records could become out of sync; individual records have their expiration set once when initially saved, but the list of unviewed ids has its expiration renewed every time a record is added to the list. The result is that the records would expire, but the list would continue to reference those records.

To ensure that `RedisStore#get_unviewed_ids` only returns ids with matching existing records, it now iterates through the list of unviewed ids and removes members that do not exist.

A test has been added to `redis_store_spec.rb` to expose this issue. Additional cleanup work has been performed in this spec; notably, the store db is now flushed prior to each spec. This revealed some broken tests that I've attempted to clean up and fix.

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -334,9 +334,9 @@ module Rack
 
 
       begin
+        @storage.save(page_struct)
         # no matter what it is, it should be unviewed, otherwise we will miss POST
         @storage.set_unviewed(page_struct[:user], page_struct[:id])
-        @storage.save(page_struct)
 
         # inject headers, script
         if status >= 200 && status < 300
--- a/lib/mini_profiler/storage/redis_store.rb
+++ b/lib/mini_profiler/storage/redis_store.rb
@@ -14,11 +14,11 @@ module Rack
       end
 
       def save(page_struct)
-        redis.setex "#{@prefix}#{page_struct[:id]}", @expires_in_seconds, Marshal::dump(page_struct)
+        redis.setex prefixed_id(page_struct[:id]), @expires_in_seconds, Marshal::dump(page_struct)
       end
 
       def load(id)
-        key = "#{@prefix}#{id}"
+        key = prefixed_id(id)
         raw = redis.get key
         begin
           Marshal::load(raw) if raw
@@ -31,24 +31,34 @@ module Rack
 
       def set_unviewed(user, id)
         key = user_key(user)
-        redis.sadd key, id
-        redis.expire key, @expires_in_seconds
+        if redis.exists(prefixed_id(id))
+          expire_at = Time.now.to_i + redis.ttl(prefixed_id(id))
+          redis.zadd(key, expire_at, id)
+        end
+        redis.expire(key, @expires_in_seconds)
       end
 
       def set_all_unviewed(user, ids)
         key = user_key(user)
-        redis.del key
-        ids.each { |id| redis.sadd(key, id) }
-        redis.expire key, @expires_in_seconds
+        redis.del(key)
+        ids.each do |id|
+          if redis.exists(prefixed_id(id))
+            expire_at = Time.now.to_i + redis.ttl(prefixed_id(id))
+            redis.zadd(key, expire_at, id)
+          end
+        end
+        redis.expire(key, @expires_in_seconds)
       end
 
       def set_viewed(user, id)
-        redis.srem user_key(user), id
+        redis.zrem(user_key(user), id)
       end
 
+      # Remove expired ids from the unviewed sorted set and return the remaining ids
       def get_unviewed_ids(user)
-        remove_expired_ids(user)
-        redis.smembers(user_key(user))
+        key = user_key(user)
+        redis.zremrangebyscore(key, '-inf', Time.now.to_i)
+        redis.zrevrangebyscore(key, '+inf', '-inf')
       end
 
       def diagnostics(user)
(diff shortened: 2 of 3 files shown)


SamSaffron:
I am very happy with this change, but are there any performance considerations here?

sayhiben:
@SamSaffron Clearing expired ids from that list certainly isn't free. Here's the code you should be concerned about, if any:

```
  def remove_expired_ids(user)
    key = user_key(user)
    redis.smembers(key).each do |id|
      redis.srem(key, id) unless redis.exists("#{@prefix}#{id}")
    end
  end
```

So, it loops through all members in `key`: O(n)
Then it does an `exists` call: O(1)
Then, if that exists call fails, it calls `srem`: O(n)

At worst, this sounds like O(n^2) if every key is expired. However, the next time `#get_unviewed_ids` runs, it will be much closer to O(n).

SamSaffron:
Maybe we should do this in lua? that way its ultra cheap always and does
not roundtrip to redis that much.

I am 100% fine to require "redis supporting lua" for rack-mini-profiler,
lua has been around for a real long time.

sayhiben:
@SamSaffron It's been a long time since I've written any lua, and I've never done so in the context of redis. Does this meet your expectations? If so, I'll add it to this PR.

```
def remove_expired_ids(user)
  key = user_key(user)
  redis.eval <<-LUA
    for _, id in pairs(redis.call('smembers', '#{key}')) do 
      if 0 == redis.call('exists', '#{@prefix}' .. id) then 
        redis.call('srem', '#{key}', id) 
      end 
    end
  LUA
end
```

SamSaffron:
Looks good to me

SamSaffron:
@sayhiben I want to merge this in, just want you to confirm that the removal logic runs super fast. 

Simplest way to do so is to use this on a live site and then add something like this to the method. 

```
t = Time.new
... Code for removal logic ... 
puts "#{Time.new - t}"
```

My gut tells me this code is super fast but I would like to confirm.

sayhiben:
@SamSaffron The data I have at the moment isn't suitable for benchmarking, so I wrote some tests that call `RedisStore#set_unviewed` with random ids a number of times, then runs the benchmarked `RedisStore#get_unviewed_ids` that executes the removal logic. 

Here's the testing code:
```ruby
    it 'tests 1000' do
      populate(user, 1000) # Simulates N expired ids in the unviewed id set
      store.get_unviewed_ids(user) # Executes benchmarked RedisStore#remove_expired_ids
    end

    def populate(user, n)
      puts "Set size: #{n}"
      n.times do |_i|
        store.set_unviewed(user, SecureRandom.hex(10))
      end
    end

  # ---

      def remove_expired_ids(user)
        t = Time.new
        key = user_key(user)
        redis.eval <<-LUA
          for _, id in pairs(redis.call('smembers', '#{key}')) do 
            if 0 == redis.call('exists', '#{@prefix}' .. id) then 
              redis.call('srem', '#{key}', id) 
            end 
          end
        LUA
        puts "#{Time.new - t} seconds"
      end
```

Here are the results after averaging over 5 runs:
```
Set size: 10
0.0002614484 seconds

Set size: 100
0.0014414862 seconds

Set size: 1000
0.0043277304 seconds

Set size: 10000
0.0441353184 seconds

Set size: 100000
0.5774553992 seconds
```
I would be surprised to learn that consumers of this library regularly have anywhere near 100,000 expired records in their store.

SamSaffron:
good point, though, if you happen to have 100k non-expired results would it
not have the same impact (0.5 delay on every new page render) ?

sayhiben:
Results are similar for non-expired sets over a 5 run average:

```
10:     0.000593901 seconds
100:    0.0015910758 seconds
1000:   0.0029252442 seconds
10000:  0.0575419732 seconds
100000: 0.377700986 seconds
```

I find this somewhat surprising, as skipping the call to `srem` should drop this from O(n^2) to O(n). However, also keep in mind that `includes.js` makes a request for each id returned, so although this solution may account for an additional 0.5s request time in extreme cases, this time will likely be dwarfed by the performance impact of making 100k requests to /mini-profiler/results.

There's potentially a faster solution that skips the lua call by switching to a sorted set internally to RedisStore. If the individual records in the set of unviewed ids are scored by their corresponding expiration timestamps, we could simply call [ZREMRANGEBYSCORE](https://redis.io/commands/zremrangebyscore) with the current time to cull expired ids.

As for potential impact of this alternative, it would make the following changes:
- `RedisStore#set_unviewed` and `RedisStore#set_all_unviewed` make calls to `Redis#sadd`, which is O(1). This would change to `Redis#zadd`, which is O(N).
- `RedisStore#set_viewed` makes a call to `Redis#srem`, which is O(N). This would change to `Redis#zrem`, which is O(M*log(N)) with N being the number of elements in the sorted set and M the number of elements to be removed.
- The call to `Redis#smembers` in `RedisStore#get_unviewed_ids` would go from O(N) to `Redis#zrange`, which is O(log(N)+M) with N being the number of elements in the sorted set and M the number of elements returned.
- `RedisStore#remove_expired_ids` would replace the O(n^2) lua call with a call to `Redis#zremrangebyscore`, which is O(log(N) + M) with N being the number of elements in the sorted set and M the number of elements removed by the operation

One important advantage to this alternative is that it would ensure consistent ordering. Currently, it appears possible to return results in an incorrect order.

SamSaffron:
Yeah using a sorted set here sounds like a great idea, and should be way
cleaner. Maybe give it a shot?

sayhiben:
@SamSaffron Calling ZREMRANGEBYSCORE has drastically improved the performance of syncing the unviewed ids. It's now taking 1/10th the time for 100k expired records. Take a look at the latest commit and let me know if you have further changes you'd like to see.

SamSaffron:
to me the implementation looks correct and low impact 

thanks heaps 👍

PR #293, opened 2017-05-19
SQL query matcher should work on "<model>.<attr>".

dcluna:
The following query broke the code inside the `match`:

```
SELECT  1 AS one FROM "my_model" WHERE LOWER("my_model"."name") = LOWER($1) AND ("my_model"."id" != $2) LIMIT $3
```

When tracing the execution, it seems to me that the regex inside `match` should
match the combination of `model_name` and `attribute_name`, which was fixed by
adding the dot to the whitespace at the end of the regex.

---

I couldn't find the specs for this, nor I'm sure about the long-term consequences of this change, but I'd be happy to work with the maintainers to solve both issues.

### DIFF
--- a/spec/lib/patches/db/pg_spec.rb
+++ b/spec/lib/patches/db/pg_spec.rb
@@ -1,0 +1,21 @@
+require 'pg'
+
+require './lib/patches/db/pg'
+
+require 'spec_helper'
+
+describe PG::Connection do
+  let(:conn) { described_class.open(dbname: 'test') }
+
+  describe '.get_binds' do
+    let(:params) { ['id', 'name'] }
+
+    let(:sql) { 'SELECT  1 AS one FROM "my_model" WHERE LOWER("my_model"."name") = LOWER($1) AND ("my_model"."id" != $2) LIMIT $3' }
+
+    let(:args) { [sql, params] }
+
+    it 'should work' do
+      expect { conn.get_binds(args) }.not_to raise_error
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'http://rubygems.org'
 gemspec
 
 gem 'codecov', :require => false, :group => :test
+gem 'pg'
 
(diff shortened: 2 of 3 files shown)


SamSaffron:
oh my ... we got to add some tests here thanks for this.

SamSaffron:
do you feel like adding a few specs for the bind stuff?

dcluna:
@SamSaffron I don't mind at all, but I'll do so over the course of this weekend. Thanks for being so quick to answer!

spec/lib/patches/db/pg_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'pg'
```
dcluna: This breaks the `SQLPatches` spec, because it'll only patch the `pg`-related code. @SamSaffron can you advise?

spec/lib/patches/db/pg_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+require './lib/patches/db/pg'
+
+require 'spec_helper'
+
+describe PG::Connection do
+  let(:conn) { described_class.open(dbname: 'test') }
```
dcluna: This is a custom local database, made just to test this. @SamSaffron can you think of a better way of doing this, or should I write some set-up code for this db in case it's not been created?

SamSaffron: oh yeah ... this is a nasty problem ... maybe just instead of making a DB connections here we can pass it various strings and binds and see that it behaves in a sane way?

SamSaffron:
OK, I reverted the binding stuff completely now, if you feel like working on this and see this feature as valuable I would love a PR to correct and test the implementation but I don't feel it is safe to ship the feature as is. Completely my fault, did a release to revert it.

nateberkopec:
Closing as stale (and the original bug was reverted). The code changed here no longer works this way, so I'm not sure this is a problem anymore.

PR #294, opened 2017-05-22
Fix `get_binds` on pg

rono23:
I think `get_binds` is a little buggy now. This PR will fix one of problems that a query has something like `LOWER()`.

```
pry(main)> PG::Connection.new.get_binds("SELECT * FROM \"posts\" WHERE LOWER(\"posts\".\"slug\") = LOWER($1) AND \"posts\".\"user_id\" = $2 LIMIT $3", ["test", 1, 2])

# Before
NoMethodError: undefined method `split' for nil:NilClass

# After
=> {nil=>"test", "user_id"=>1, "LIMIT"=>2}
```

And there is another problem that [this regular expression](https://github.com/MiniProfiler/rack-mini-profiler/blob/v0.10.4/lib/patches/db/pg.rb#L118) does not care about `LOWER()`, `!=` etc, so `get_binds` returns a hash with `nil` key etc (this PR does not include this fix).

```
pry(main)> PG::Connection.new.get_binds("SELECT * FROM \"posts\" WHERE LOWER(\"posts\".\"slug\") = LOWER($1) AND \"posts\".\"user_id\" = $2 LIMIT $3", ["test", 1, 2])
=> {nil=>"test", "user_id"=>1, "LIMIT"=>2}

pry(main)> PG::Connection.new.get_binds("SELECT * FROM \"posts\" WHERE LOWER(\"posts\".\"slug\") = LOWER($1) AND \"posts\".\"user_id\" != $2 LIMIT $3", ["test", 1, 2])
=> {nil=>"test", "!="=>1, "LIMIT"=>2}
```

### DIFF
--- a/lib/patches/db/pg.rb
+++ b/lib/patches/db/pg.rb
@@ -103,24 +103,23 @@ class PG::Connection
     return if params.nil? || params.empty?
 
     sql = args[0]
-
-    hh = {}
     arr = nil
 
     if sql.match(/\(\$\d/)
-      arr = sql.match(/\((\"\w+\",?\s?)+\)/).to_s.gsub!(/[",()]/, '').split #regular selects list of names from sql query, for ex. ("name1", "name2", "name3")
+      # regular selects list of names from sql query, for ex. ("name1", "name2", "name3")
+      arr = sql.match(/\((\"\w+\",?\s?)+\)/).to_s.gsub!(/[",()]/, '').to_s.split
     end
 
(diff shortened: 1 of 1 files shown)


rono23:
Oh.. sorry I didn't check #293. Please close this if you don't need.

SamSaffron:
Sorry, going to close this for now, the PR I merged was not properly tested, I just did a new release reverting all the bind sniffing stuff.

PR #295, opened 2017-05-23
[ci skip] Document the need for Redis in Prod

schneems:
While we document the `RedisStore` I want to explicitly call out that it is required when setting up a "non-development environment".

Minor typos and formatting clean up.


SamSaffron:
cool!!! 👍

Issue #262, opened 2016-08-20
Position bottom (left/right)?


mrkongo:
Maybe is necessary another config option for bottom positioning?

SamSaffron:
sure, feel free to submit a PR for custom positioning

rafaelbiriba:
Waiting for review and approval 😄 
Please do it as soon as possible, since I need to use this gem in "bottom" position.
Thanks 🙏

rafaelbiriba:
@SamSaffron you can close this ticket!

SamSaffron:
thanks 👍

PR #298, opened 2017-06-07
Add support for vertical_position (top/bottom) parameter.

rafaelbiriba:
This PR closes the issue (#262)


SamSaffron:
Cool! can I see some screenshots? Also can you add a changelog entry

rafaelbiriba:
I can not show you the sql print because it have sensitive data, but it's working.
I tested: "left+top", "left+bottom", "right+top", "right+bottom" 
screen shot 2017-06-07 at 15 04 36

SamSaffron:
cool, what happens if there are multiple reqs, how does it "pile" up?

rafaelbiriba:
vertically, I tested that as well, hahaha
and also the rounded corner stays on the first element (if bottom is enabled).
I am justing wait the new gem version to put the code in production. (It's is an internal admin system)

SamSaffron:
Actually I think we should refine this a tiny bit... let's just keep the position arg, and extend it

`right` == `top-right` (also allowed)
`left` == `top-left` (also allowed) 
`bottom-right` (new)
`bottom-left` (new) 

It means there is one less argument which is good.

rafaelbiriba:
Yes, but with an expensive trade-off.
A lot of places in JS we use the `.position` directly to position the element.
Accepting something like `left` and `bottom-right` means that we need to parse this information everytime, checking if the argument is simple `left` or with dash `bottom-left`.

SamSaffron:
not that expensive, you would only need to parse that string once on the client up front and set a few vars. I don't see this as such a major change on top of your PR.

rafaelbiriba:
Take another look!

Print about the stack report.
screen shot 2017-06-07 at 16 50 05

lib/mini_profiler/profiler.rb
```diff
@@ -586,19 +586,22 @@ def ids_comma_separated(env)
     def get_profile_script(env)
       path = "#{env['RACK_MINI_PROFILER_ORIGINAL_SCRIPT_NAME']}#{@config.base_url_path}"
 
+      position = @config.position.match("-") ? @config.position.split("-") : ["top", @config.position]
```
SamSaffron: I think its way cleaner just to have Rack::MiniProfiler::Config handle this. A lot easier to test and can amend vertical / horizontal when you set position.

rafaelbiriba:
WDYT?

lib/mini_profiler/profiler.rb
```diff
@@ -67,6 +67,7 @@ def request_authorized?
     # :auto_inject - should script be automatically injected on every html page (not xhr)
     def initialize(app, config = nil)
       MiniProfiler.config.merge!(config)
+      MiniProfiler.config.set_positions!
```
SamSaffron: I don't think we need this #merge! can take care of it, just move logic there.

lib/mini_profiler/config.rb
```diff
@@ -81,6 +80,11 @@ def merge!(config)
         end
       end
 
+      def set_positions!
+        position = (self.position && self.position.match("-")) ? self.position.split("-") : ["top", "left"]
+        instance_variable_set "@vertical_position", position.first
```
SamSaffron: this is weird, why use instance_variable_set? just set defaults in the constructor and then `@vertical_position =`

rafaelbiriba:
`set_position` moved to inside merge (after merge any possible configuration).

about instance_variable_set, I change it to self.attribute, since I am inside the class.

about the constructor, the position needed to be calculated after constructor block, because I need to check and split the value (if applicable).

SamSaffron:
Cool, lets try this, thanks for :bear: ing with me !

PR #302, opened 2017-06-22
Change the railtie to allow other stores than the file store

### LINKED ISSUE #301: Railtie prevents any type of storage other than file storage
The railtie as it's set up forces you to use the FileStore storage option due to [this line](https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler_rails/railtie.rb#L44).

This makes it impossible to follow any of the directions from the readme around changing your storage option and it had me scratching my head for a long time 😀
So what I'm seeing is that you're not able to change the storage beforehand because the railtie will overwrite whatever storage option you see and you're not able to change the storage after the profiler is initialized because it stores an instance of the storage class at the time of initialization so you wind up being stuck!

I'm not sure the best way to resolve this. You have a comment in the railtie about the file store being better than the default memory store so maybe it would work to change the default from the memory store to the file store? I'm open to any suggestions on a way to resolve this.

Currently I just copied the contents of the railtie initializer to my local project and took out the part where it assigns the file store so I'm working for now but it would be great if we could fix this so other people aren't confused.

Thanks

willtcarey:
Check to see if the config options are still at the default and if they are then go ahead and use the file store. This will not cause any changes for people who are already using the gem. If the config options are different than the defaults then leave them as they are to allow the instructions in the readme to work.

Fixes #301

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -37,11 +37,14 @@ module Rack::MiniProfilerRails
     end
 
     # The file store is just so much less flaky
-    base_path = Rails.application.config.paths['tmp'].first rescue "#{Rails.root}/tmp"
-    tmp       = base_path + '/miniprofiler'
+    # If the user has not changed from the default memory store then switch to the file store, otherwise keep what the user set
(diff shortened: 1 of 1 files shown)


SamSaffron:
looks good to me! thanks

PR #303, opened 2017-06-28
Freeze some trivial common strings

viraptor:
Making sure these are frozen saved around half a million allocations per
request in a large app.

### DIFF
--- a/lib/mini_profiler/timer_struct/sql.rb
+++ b/lib/mini_profiler/timer_struct/sql.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module Rack
   class MiniProfiler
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
my preference here is actually to move files to use `# frozen_string_literal: true` do you care to revise with that comment on this file? 

I really only care about this kind of saving in Ruby 2.3 and up anyway.

viraptor:
Done. This makes `< 2.3` a bit worse because of the dup, but does improve `>= 2.3`

SamSaffron:
Why not use `String.new` there instead?

SamSaffron:
Also, if you feel like working through more files, please do! I would love it!

viraptor:
Didn't think of that :) Updated.

SamSaffron:
Thanks heaps!

PR #304, opened 2017-06-30
Add "X-MiniProfiler-Original-Cache-Control" to store original Cache-Control

mrasu:
Make store original "Cache-Control" of header.

Because I'm using CDN and need to change "Cache-Control" header, it is important to know the real header in production


SamSaffron:
sure, seems reasonable, thanks

PR #305, opened 2017-07-17
InstallGenerator for Rails application

yhirano55:
I've implemented RackProfiler::Generators::InstallGenerator to be easy to set up rack-mini-profiler  for Rails application:

```bash
$ bin/rails g rack_profiler:install
```

- The filename of initializers is named not `rack-mini-profiler.rb` but `rack_profile.rb`. Becase filename is expected to be not kebab-case but snake-case.
- Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,15 +63,10 @@ gem 'rack-mini-profiler', require: false
 Note the `require: false` part - if omitted, it will cause the Railtie for the mini-profiler to
 be loaded outright, and an attempt to re-initialize it manually will raise an exception.
 
-Then put initialize code in file like `config/initializers/rack_profiler.rb`
+Then run the generator which will set up rack-mini-profiler in development:
 
-```ruby
-if Rails.env == 'development'
-  require 'rack-mini-profiler'
-
-  # initialization is skipped so trigger it
-  Rack::MiniProfilerRails.initialize!(Rails.application)
-end
+```bash
+bundle exec rails g rack_profiler:install
 ```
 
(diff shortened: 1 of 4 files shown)


SamSaffron:
OK, this seems fine to me, thanks!

krainboltgreene:
Is the generator gone? The latest install says no generator exists by that name.

yhirano55:
@krainboltgreene Hi. It seems this generator have not released on [v0.10.5](https://github.com/MiniProfiler/rack-mini-profiler/tree/v0.10.5/lib), so if you want to use this, you need to fetch from master branch on GitHub.

ksykulev:
When using the `FileStore`, upgrading from 0.10.5 -> 0.10.6 with the old initializer 
```
#config/initializers/rack_profiler.rb
Rack::MiniProfiler.config.storage = Rack::MiniProfiler::FileStore if defined?(Rack::MiniProfiler)
```

yields an error:
```
gems/rack-mini-profiler-0.10.6/lib/mini_profiler/storage/file_store.rb:47:in `initialize': path (ArgumentError)
```

In order to fix this you need to switch to the new initializer syntax
```
if Rails.env.development?
  require "rack-mini-profiler"

  # initialization is skipped so trigger it
  Rack::MiniProfilerRails.initialize!(Rails.application)
  Rack::MiniProfiler.config.storage = Rack::MiniProfiler::FileStore if defined?(Rack::MiniProfiler)
end
```

PR #306, opened 2017-08-03
Upgrade to Rspec 3

### LINKED ISSUE #219: RSpec 3
As I discovered in #215 / #217, we're still using RSpec 2. The big change is a syntactic one:

``` ruby
# RSpec 2
author.name.should == 'Julie'

# RSpec 3
expect(author.name).to eq 'Julie'
```

While the new syntax is more to read and more to type, I feel that it's better because it means RSpec embeds less methods onto all objects.

@SamSaffron what do you think? I don't want to go to the trouble of creating a PR for this if you prefer the old syntax.

rrosenblum:
This fixes #219. I tried setting the test order to random, but there were a few test failures. I am interested in learning more about the inner workings of rack-mini-profiler so I may try to fix that soon.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,2 @@
+--color
+--require spec_helper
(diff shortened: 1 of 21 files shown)


SamSaffron:
sorry for the huge delay here, we are all merged now!

SamSaffron:
❤️

PR #308, opened 2017-08-21
Fix toggle shortcut doesn't work

lukesilvia:
Hello. 
Toggle shortcut doesn't work on `v0.10.5`.

Keydown event for the shorcut isn't added due to wrong syntax.
Second argument `null` is required. (refs: [README](https://github.com/jeresig/jquery.hotkeys/blob/0d75b7907961d9f60ad1e3bb82f146cc7a3a8996/README.md#jqueryhotkeys-) )

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -400,7 +400,7 @@ var MiniProfiler = (function () {
                 popupHide(button, popup);
             }
         });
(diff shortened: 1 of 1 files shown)


SamSaffron:
seems fine, thanks!

PR #312, opened 2017-09-29
Suppress profiler results in print media

md5:


SamSaffron:
looks good to me! thanks

PR #315, opened 2017-10-23
Record query plan for postgres SELECT statements.

mwulfvhl:
Seeing the query plan generated by EXPLAIN SELECT is useful for my current project. Here it's recorded only by the PG patch; it could be implemented in other patches as well.

NOTE: getting this ready to be considered for actual merge will be a longer-term project.

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -53,12 +53,14 @@ var MiniProfiler = (function () {
         var key = 'templates',
             cached = load(key);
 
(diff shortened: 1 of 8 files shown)


SamSaffron:
I am leaning towards allowing this kind of thing but clearly it will need

- Either to be an optional default off
- OR, invoked via the UI after the fact

nateberkopec:
Closing as stale.

PR #316, opened 2017-10-23
Fix error with webrick and empty cache control

jsteel:
Webrick assumes each header key has a value. Don't store the
old Cache-Control value unless it existed.

The error caused by Webrick is the same as the one detailed here: https://mensfeld.pl/2013/01/ruby-on-rails-webrick-error-nomethoderror-undefined-method-split-for-nilnilclass/

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -366,7 +366,7 @@ module Rack
         headers.delete('Date')
       end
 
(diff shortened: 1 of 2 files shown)


jsteel:
The Travis failures appear to have nothing to do with these changes.

ttwo32:
Good patch. 👍 
I am faced with a same situation.

SamSaffron:
Cool, thanks for this seems good to me!

vochicong:
Thank you!

PR #318, opened 2017-10-30
Add source code and changelog links to gemspec

greysteil:
Makes it easy to programatically find the source code and changelog for `rack-mini-profiler`, using the rubygems API. More details on the direction of travel for gemspecs is [here](https://github.com/rubygems/rubygems.org/issues/1127) and [here](https://github.com/rubygems/rubygems.org/pull/1234), and the current state is [here](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification.rb).

(My interest in this is so that [Dependabot](https://dependabot.com) can find a link to the `rack-mini-profiler` source code when it creates PRs.)


SamSaffron:
looks good!

PR #319, opened 2017-10-30
Added frozen string literal in all files

redtachyons:
Enable frozen string literal inn all files

- [x] Make all tests passing
- [x] Fix things which are not covered in tests
- [x] Readme fix
- [ ] Changelog

### DIFF
--- a/lib/mini_profiler/timer_struct/client.rb
+++ b/lib/mini_profiler/timer_struct/client.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module Rack
   class MiniProfiler
     module TimerStruct
@@ -16,10 +17,7 @@ module Rack
         # used by Railtie to instrument asset_tag for JS / CSS
         def self.instrument(name, orig)
           probe = "<script>mPt.probe('#{name}')</script>"
-          wrapped = probe
-          wrapped << orig
-          wrapped << probe
-          wrapped
+          probe + orig + probe + orig
         end
 
 
--- a/lib/patches/db/nobrainer.rb
+++ b/lib/patches/db/nobrainer.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 class Rack::MiniProfiler::NoBrainerProfiler
 
   def on_query(env)
--- a/lib/patches/db/nobrainer.rb
+++ b/lib/patches/db/nobrainer.rb
@@ -10,16 +10,16 @@ class Rack::MiniProfiler::NoBrainerProfiler
       query = ""
 
       # per-model/query database overrides
-      query << "[#{env[:options][:db]}] " if env[:options][:db]
+      query += "[#{env[:options][:db]}] " if env[:options][:db]
 
       # "read", "write" prefix
       # query << "(#{NoBrainer::RQL.type_of(env[:query]).to_s}) "
 
-      query << "NOT USING INDEX: " if not_indexed
-      query << env[:query].inspect.gsub(/\n/, '').gsub(/ +/, ' ') + " "
+      query += "NOT USING INDEX: " if not_indexed
+      query += env[:query].inspect.gsub(/\n/, '').gsub(/ +/, ' ') + " "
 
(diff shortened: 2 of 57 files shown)


SamSaffron:
❤️ 💟 😍  thanks heaps for working on this!

cbliard:
Regarding the proposed changes, wouldn't it be easier and more performant to use `String.new`and keep using `<<` instead of replacing them with `+=`.

For instance from this piece of code:

```
body = "Rack Environment\n---------------\n"
env.each do |k,v|
  body << "#{k}: #{v}\n"
end
body << "\n\nEnvironment\n---------------\n"
text_result(body)
```

Instead of changing it to this:
```
body = "Rack Environment\n---------------\n"
env.each do |k,v|
  body += "#{k}: #{v}\n"
end
body += "\n\nEnvironment\n---------------\n"
text_result(body)
```

Use this:
```
body = String.new("Rack Environment\n---------------\n")
env.each do |k,v|
  body << "#{k}: #{v}\n"
end
body << "\n\nEnvironment\n---------------\n"
text_result(body)
```

redtachyons:
I thought about the same, But in that case, we are keeping strings as mutable, right? 

> and more performant to use 

Can you please share the performance benchmarks comparing String.new and immutable string operations

cbliard:
Exactly, some strings are kept as mutable and the majority are frozen. Instead of having all strings mutables by default and choosing which ones are frozen, the frozen string literal option makes all strings frozen by default and we choose which ones are mutable. Having mutable strings is not forbidden.

About benchmarking `+=` and `<<`, here is a little benchmark using the `benchmark-ips` gem

```
require 'benchmark/ips'

CHAR = 'A'.freeze
[10, 100, 1000, 10000].each do |count|
  puts "concatenating #{count} strings together"
  Benchmark.ips do |x|
    x.config(time: 3, warmup: 1)
    x.report("+=") do
      s = ""
      count.times { s += CHAR }
    end

    x.report("<<") do |times|
      s = String.new
      count.times { s << CHAR }
    end

    x.compare!
  end
end
```

It shows that `+=` is an order of magnitude slower than `<<`, which is logical because += allocates a new string everytime and that creates a lot of short-living objects.

I have not read mini-profiler code thoroughly and this may be a micro-optimization as it does not matter so much if there are only a couple of string concatenations.

SamSaffron:
Huge warning on string.new, you get an ascii encoded string back, always
prefer “”.dup which is utf8

cbliard:
Thanks, I did not know about that.

lib/patches/db/nobrainer.rb
```diff
@@ -9,16 +10,16 @@ def on_query(env)
       query = ""
```
SamSaffron: The general pattern should be ... 

```
query = "".dup
```

Then rest of the code is unchanged, all the string copying here should probably change

SamSaffron:
@cbliard  agree, the general pattern is that any "mutable" string we have should be initialized with `"".dup` then later on we can still use `<<` like we do today.

lib/mini_profiler/timer_struct/client.rb
```diff
@@ -16,10 +17,7 @@ def self.init_instrumentation
         # used by Railtie to instrument asset_tag for JS / CSS
         def self.instrument(name, orig)
           probe = "<script>mPt.probe('#{name}')</script>"
-          wrapped = probe
-          wrapped << orig
-          wrapped << probe
-          wrapped
+          probe + orig + probe + orig
```
cbliard: Warning here, there is an extra `orig`. It should be `probe + orig + probe`.

redtachyons: check the test here https://github.com/MiniProfiler/rack-mini-profiler/blob/master/spec/lib/timer_struct/client_timer_struct_spec.rb#L41

cbliard: Indeed. Sorry for the burden.

PR #320, opened 2017-11-10
Replace Time.now with Process.clock_gettime(Process::CLOCK_MONOTONIC) everywhere

jamescook:
Basically just a search and replace along with a new helper that replaces `Time.travel`. I also fixed an unrelated test failure. 

I see there's another PR that updates the README in a similar fashion, so that will probably be a conflict.


SamSaffron:
Awesome this looks great to me!

PR #324, opened 2017-12-01
Add `# frozen_string_literal: true` to all `lib/**/*.rb` files

ccmeyers:
Knowing that there is another PR out for this fix (https://github.com/MiniProfiler/rack-mini-profiler/pull/319), feel free to deprioritize this fix.

A few notable differences:
1. Used `.dup` instead of switching `<<` to `+=` to cut down on diff
2. Did not add `# frozen_string_literal: true` to spec files (was this something that was wanted?)

Also noticed that there were a few strings that should have failed, but didn't due to lack of coverage around those methods. We found this in:
1. `execute_with_profiling` in `lib/patches/db/rsolr.rb:12`
2. ~~`analyze_memory` in `lib/mini_profiler/profiler.rb:480`~~
3. `on_query` in `lib/patches/db/nobrainer.rb:11`
4. ~~`initialize` in `lib/mini_profiler/timer_struct/sql.rb:14`~~ I was wrong, this was covered

Thank you so much for maintaining this awesome gem!

### DIFF
--- a/lib/mini_profiler/timer_struct/sql.rb
+++ b/lib/mini_profiler/timer_struct/sql.rb
@@ -11,7 +11,7 @@ module Rack
           stack_trace = nil
           unless skip_backtrace || duration_ms < Rack::MiniProfiler.config.backtrace_threshold_ms
             # Allow us to filter the stack trace
-            stack_trace = String.new.dup
+            stack_trace = "".dup
              # Clean up the stack trace if there are options to do so
             Kernel.caller.each do |ln|
               ln.gsub!(Rack::MiniProfiler.config.backtrace_remove, '') if Rack::MiniProfiler.config.backtrace_remove and !full_backtrace
--- a/lib/mini_profiler/timer_struct/sql.rb
+++ b/lib/mini_profiler/timer_struct/sql.rb
@@ -11,7 +11,7 @@ module Rack
           stack_trace = nil
           unless skip_backtrace || duration_ms < Rack::MiniProfiler.config.backtrace_threshold_ms
             # Allow us to filter the stack trace
-            stack_trace = String.new
+            stack_trace = String.new.dup
              # Clean up the stack trace if there are options to do so
             Kernel.caller.each do |ln|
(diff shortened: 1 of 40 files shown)


SamSaffron:
Thanks so much for picking this up!

lib/mini_profiler/timer_struct/sql.rb
```diff
@@ -10,7 +11,7 @@ def initialize(query, duration_ms, page, parent, params = nil, skip_backtrace =
           stack_trace = nil
           unless skip_backtrace || duration_ms < Rack::MiniProfiler.config.backtrace_threshold_ms
             # Allow us to filter the stack trace
-            stack_trace = String.new
+            stack_trace = String.new.dup
```
SamSaffron: we should change this to `"".dup` instead of `String.new.dup`

SamSaffron:
Awesome, happy to merge this in! would you be OK to add some missing tests to catch any of the uncovered areas?

ccmeyers:
Sounds good. I'll take a look at the specs.

SamSaffron:
@donv its a bit tricky, we got to support all stable versions of Ruby (some people even want 1.9 which I kind of object to) I think that syntax is quite new. 

@ccmeyers do you think this is ready to be merged in?

ccmeyers:
@SamSaffron almost, I'm just having trouble figuring out coverage for two of the places I mentioned in the description.

donv:
@SamSaffron Unary plus was present in Ruby 1.9.2 at least.

donv:
@SamSaffron OK, looks like the semantics for unary plus on String was added in Ruby 2.3.  Not a good idea to use it while Ruby 2.2 (and older) is supported then.

SamSaffron:
k, I am merging this now, we can refine in future PRs, thanks heaps @ccmeyers, enjoy the holiday season!

ccmeyers:
Thanks @SamSaffron! Happy Holidays!

PR #330, opened 2018-01-29
Drop TODO ancient comment

olleolleolle:
This PR just drops an undone TODO comment which refers to code no longer in the repository, if it ever were here.


SamSaffron:
thanks!

PR #331, opened 2018-02-02
Ruby Warnings

amatsuda:
Here's an attempt to fix some Ruby-level warnings that we see with `-w` Ruby option.


SamSaffron:
Awesome, thanks heaps @amatsuda  !

PR #332, opened 2018-02-07
Store profiler visibility preference for a session

MikeRogers0:
I thought it would be neat when a user presses `alt + p`, it stores the profiler visibility preference for their entire session instead of just on that page load.

What do you think?

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -402,6 +402,7 @@ var MiniProfiler = (function () {
         });
         $(document).on('keydown.mini-profiler', null, options.toggleShortcut, function(e) {
             $('.profiler-results').toggle();
+            sessionStorage['rack-mini-profiler-start-hidden'] = $('.profiler-results').is(':hidden');
         });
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am fine with this, but in some cases people have this disabled so we can not have access to it crash... all access to it needs to be guarded. (unable to write due to exception = no-op, unable to read due to it missing from window or something = no-op) 

I can't remember exactly how you get browsers to break this, but possibly private browsing mode in safari? not sure.

MikeRogers0:
> I can't remember exactly how you get browsers to break this, but possibly private browsing mode in safari?

I'll try to break it, then report back :)

MikeRogers0:
I wasn't able to recreate an issue with the `alt + p` visibility preference being stored in the `sessionStore` in Safari, in private mode 👍

tomkadwill:
+1 for this PR

SamSaffron:
Ok let's try this out next release, thanks!

MikeRogers0:
Oh awesome!!! Thank you!

PR #333, opened 2018-02-11
Fix page start time

superiorlu:
hi, i  find the profiler ui on `includes.tmpl`, the start time is not correct, it use `Process.clock_gettime` for initialize `Date`, so i  do that:
``` ruby
# page.rb
def initialize
  ...
  started_at   = (Time.now.to_f * 1000).to_i
  started      = (Process.clock_gettime(Process::CLOCK_MONOTONIC) * 1000).to_i
  ...
end

...

def extra_json
  ....
  :started   => '/Date(%d)/' % @attributes.delete(:started_at)
  ... 
end
```
i hope that have a suggestion from you soon. thanks.

### DIFF
--- a/lib/mini_profiler/timer_struct/page.rb
+++ b/lib/mini_profiler/timer_struct/page.rb
@@ -13,12 +13,14 @@ module Rack
         def initialize(env)
           timer_id     = MiniProfiler.generate_id
           page_name    = env['PATH_INFO']
-          started_at   = (Process.clock_gettime(Process::CLOCK_MONOTONIC) * 1000).to_i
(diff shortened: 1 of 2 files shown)


SamSaffron:
Yeah I think this is right, but where is started being used then? If this is just for the UI then no point carrying around.

superiorlu:
Yes, you are right,  i have deleted `started_at` field from json, only keep `started` field. 
The `started_at` is not only for UI, but also is a timestamp for page.

SamSaffron:
I see, its a bit of a tangled web, lets try this!

superiorlu:
@SamSaffron thanks 😄

Issue #329, opened 2018-01-29
chrome.loadTimes() is deprecated


t27duck:
Latest Chrome version (version 64) is bringing up depreciation warning regarding the use of `chrome.loadTimes()`

![chrome-dep](https://user-images.githubusercontent.com/545604/35510207-3c6ad9d0-04c5-11e8-8010-1e60687365da.png)

Relevant Chrome information page: https://www.chromestatus.com/features/5637885046816768

Relivant line in includes.js (in master): https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/html/includes.js#L113

SamSaffron:
Would love a PR to fix this.

MikeRogers0:
I'd love to open a PR for this, I think replacing https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/html/includes.js#L113-L122 with:

```
var paintMetrics = performance.getEntriesByType("paint");

if(paintMetrics !== undefined && paintMetrics.length > 0){
  paintMetrics.forEach((paintMetric)=>{
    clientPerformance.timing["paintMetric.name"] = paintMetric.startTime;
  });
}
```

SamSaffron:
looks good idea, can you give it a shot?

mathyoudawson:
Can we close this as the the PR has been merged?
@SamSaffron

SamSaffron:
sure! thanks for reminding me.

PR #335, opened 2018-02-14
Using the Web Performance API to collect stats about paint time

MikeRogers0:
Part of #329 - This fixes the deperecation warning and replaces it with
meaningful data. 

However, I think we might be able to do one better and
use the Web Performance API for all the key stats.  What do you think?

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -108,18 +108,6 @@ var MiniProfiler = (function () {
                         copy.navigation.redirectCount = clientPerformance.navigation.redirectCount;
                     }
                     clientPerformance = copy;
-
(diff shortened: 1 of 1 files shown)


SamSaffron:
excellent! thanks 

> However, I think we might be able to do one better and
use the Web Performance API for all the key stats. What do you think?

Would love that, can you try it (and maybe attach a little screen shot) there may be some extra stats you want to grab there?

MikeRogers0:
> Would love that, can you try it (and maybe attach a little screen shot) there may be some extra stats you want to grab there?

👍 I'll give it a shot & PR back :)

PR #340, opened 2018-03-19
Don't subtract top from windowHeight in popupSetDimensions

naiyt:
This was breaking the popup from working when rendered in an iframe. It would spit out an error like this:

```
VM137461:1 Uncaught DOMException: Blocked a frame with origin "http://localhost:4000" from accessing a cross-origin frame.
    at eval (eval at popupSetDimensions (http://localhost:4000/mini-profiler-resources/includes.js?v=812af6927f7534f73488ece21dc89b62:257:9), <anonymous>:1:14)
```

I believe the `maxHeight` variable wasn't working properly anyway, as `windowHeight - top - 40` return `NaN` when not in an iframe.

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -254,7 +254,7 @@ var MiniProfiler = (function () {
     var popupSetDimensions = function (button, popup) {
         var px = button.position().top - 1, // position next to the button we clicked
             windowHeight = $(window).height(),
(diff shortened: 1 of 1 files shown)


SamSaffron:
so odd, top was never even defined ... we got to jslint this file.

PR #344, opened 2018-03-26
Remove comments in README about Rubies < 2.2

nateberkopec:
No longer necessary now that gem only supports 2.2+


SamSaffron:
❤️ better readme is the best

PR #346, opened 2018-03-28
Fix jQuery 3 shorthand depraction warning for click() function

TiSer:


SamSaffron:
this looks right to me, thanks!

Issue #348, opened 2018-04-11
No bindings for clients that use the new javascript fetch API


shyam-habarakada:
I noticed that we don't have any bindings that refresh profiler results when the page is using the `fetch` API https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

Does anyone have an update for this in the works? If not I can try to get a PR going.

SamSaffron:
interesting, yeah I would like to have that added, can you make a PR

shyam-habarakada:
## design discussion

I was looking into this more — the current pattern with other libraries like jQuery and XHR appears to be to patch or hook into available extension points so that inspecting and grabbing the `X-MiniProfiler-Ids` from headers and passing that through to the `fetchResults` method.

Unfortunately the fetch API doesn't seem to have hooks that we can incorporate into and the promise based design of it makes patching it not a very attractive option. Strictly technically speaking it seems feasible to patch `fetch` and wrap the promises it returns so that we can intercept and capture the headers. Another option is to expose a method in this library that _any_ javascript could invoke to just pass in `X-MiniProfiler-Ids`. In our use of fetch for instance, we can then add something into the promise chain to do exactly that. Less intrusive but won't work by default.

Thoughts/feedback?

shyam-habarakada:
@SamSaffron PR linked above. Thanks

gregmolnar:
This one can be closed since it is fixed in #351

PR #351, opened 2018-05-19
Add support for the new javascript fetch API

shyam-habarakada:
## Proposed solution for #348. 

Add support for the new `fetch` API so we capture the `x-miniprofiler-ids` from the fetch response headers. Fetch is promise based, so we wrap it when mini profiler is enabled and setup the promise chain to first inspect it for profiling purposes and then pass on the response (or errors) to the original promise chain.

I've tested this and confirmed it works in a react app we have that is using the fetch API the mini profiler (the fact that the app is react is inconsequential to how this works).

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -603,20 +603,20 @@ var MiniProfiler = (function () {
               __originalFetch(input,init).then(function(response) {
                 // look for x-mini-profile-ids
                 for (var pair of response.headers.entries()) {
-                  if(pair[0] == 'X-MiniProfiler-Ids') {
+                  if(pair[0] && (pair[0].toLowerCase() == 'x-miniprofiler-ids')) {
                   var ids = JSON.parse(pair[1]);
                     fetchResults(ids);
                   }
                 }
-                resolve(response);  
+                resolve(response);
               }).catch(function(error) {
                 reject(error);
-              });	
+              });
             });
           }
           window.fetch.__patchedByMiniProfiler = true;
         }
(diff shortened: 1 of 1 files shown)


lib/html/includes.js
```diff
@@ -592,6 +592,31 @@ var MiniProfiler = (function () {
           };
         }
 
+        // https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
+        if(typeof(window.fetch) === 'function' && window.fetch.__patchedByMiniProfiler === undefined) {
+          var __originalFetch = window.fetch;
+
+          console.warn('alert: patching window.fetch for profiling');
```
SamSaffron: not sure about the logging though. plus some may not even have console.warn

maybe we just have an option for patching fetch if you select that then fetch is patched?

shyam-habarakada: Sure, I will exclude the warning.

Is there already a mechanism for conditionally excluding certain `included` integrations? 

My thinking was, if you are using fetch and you are using rack-mini-profiler, you want the integration. If you aren't using fetch and/or the browser does not support fetch—i.e. `window.fetch` is undefined, there is no issue.

What do you think @SamSaffron ?

shyam-habarakada:
@SamSaffron I addressed your comments. Could you have another look?

SamSaffron:
OK, lets try this out next release, last thing to add is a note in the README about what patching rack-mini-profiler does (and when), can you do a PR for that?

Issue #350, opened 2018-05-18
Can't figure out how to get a flamegraph for ajax post request.


CyborgMaster:
I can see the profiler badge for ajax requests by hitting any html page (it shows all of requests since the last html load), however I can't figure out how to get the flamegraphs for those requests.  Am I missing something?

CyborgMaster:
#352 is the way I ended up working around it for now.  

Because this is a JSON API post, there is no way to return the flamegraph directly to the browser, so the only option would be to cache it and return it on the next HTML request, in a similar fashion to the current summary stats that appear in the profile badge.  However, flamegraphs can be very large (10s or even 100s of megabytes) so caching many of these may be prohibitive.  

My solution was to add an option to save the flamegraph html to a file, where I can open it in a separate browser tab.

This is just a quick solution I threw together, but it turned out to be very effective for my use case. 
 Thoughts?

OsamaSayegh:
Perhaps I'm missing something here, but for you can append `?pp=flamegraph` to your URL and you'll get flamegraph for your request, this even works for JSON routes e.g. in any Discourse site (assuming I logged in as admin) I can do `/latest.json?pp=flamegraph` and I get a flamegraph. I think you're already aware of that, but does that not work for you or am I missing something?

CyborgMaster:
Yes, but in this case it's not a page load that is making the request, it's an Angular SPA using XHR.  Angular expects the response to be json and not html and then crashes.  Yes, I could do a manual page load instead of letting the SPA do it, but  the problem is that the slowdown only occurs in a particular sequence of API calls that the SPA is performing, so it's not feasible to have it be a full page load.

OsamaSayegh:
Ok, I think I understand your use case now: Your use case is that you want a JSON request to still return a JSON response, but you also want to somehow see the flamegraph for that request. You can't use `?pp=flamegraph` for two reasons:

1. `?pp=flamegraph` changes the response type for JSON to HTML;
2. in your application the slowdown occurs after a sequence of API calls, and if you were to use `?pp=flamegraph` there would be a quite a bit of manual work involved.

Does this sound right?

CyborgMaster:
Nailed it!  That describes the situation very well.

feliperaul:
Same here @OsamaSayegh

OsamaSayegh:
[...]

PR #352, opened 2018-05-19
Add the ability to save flamegraphs to a file

CyborgMaster:
This is useful for json api posts, and is my proposal for solving #350.

It follows the same pattern as `flamegraph_embed`, but allows you to save the flamegraph to an external html file, which you can then open in a separate browser tab.


SamSaffron:
very mixed on this, this would save it on the server and hunting down where the file is, is going to be mega difficult when deployed to a ton of web app servers

CyborgMaster:
This is true, but without this there is no way to look at them even in development, which is what my use case was.  We had a slow endpoint during development and no good way to figure out what it was locally because it was a json endpoint.  This allowed us to dive into what's going on.  I don't think we'd ever try and use it in production.

CyborgMaster:
Any further thoughts?

lib/mini_profiler/profiler.rb
```diff
@@ -331,8 +331,12 @@ def call(env)
       page_struct[:root].record_time((Process.clock_gettime(Process::CLOCK_MONOTONIC) - start) * 1000)
 
       if flamegraph
-        body.close if body.respond_to? :close
-        return client_settings.handle_cookie(self.flamegraph(flamegraph))
+        if query_string =~ /save/
+          ::File.write 'flamegraph.html', flamegraph
```
SamSaffron: I think you should write this to a file in `tmp` and then output where it was saved eg: 

```
saved flamegraph to tf23.docker  /tmp/flamegraphs_20190201-01023.html`
```

You get the hostname from `hostname`

williamcodes:
Is this PR dead? I'd love the ability to save a flamegraph. I've tried saving it in the browser but it doesn't work properly when I reopen it.

SamSaffron:
I support adding something here, but worried about where it is placing the file. if my comment is addressed I would be happy to merge.

CyborgMaster:
Sorry about that.  Somehow I missed the comment.  I'll make the changes and get it to you soon.

CyborgMaster:
@SamSaffron I saved the files in a temp directory as requested, however I wasn't sure how you wanted the filename outputted.  I can't modify the in-flight request because I don't know it's format; it could be json, xml, protobuf, etc.  In the code right now I just did `puts`, but how do you recommend it be done?

SamSaffron:
> Is this PR dead? I'd love the ability to save a flamegraph. I've tried saving it in the browser but it doesn't work properly when I reopen it.

I think the fundamental issue here is that this is already sort of supported `?pp=flamegraph_embed` allows you to get a flamegraph with all the stuff embedded so you can easily just save the html and load it in another browser / etc. 

What is the specific use case here that goes beyond the ability to share? 

I know I have flip flopped here a lot, but I worry about adding something that will be confusing to explain / use when `flamegraph_embed` already solves the issue.

williamcodes:
@SamSaffron saving the html and loading it in another browser was the first thing I tried. It doesn't save properly. I dunno if it's a css issue or a javascript issue or what. It looks like it might be okay when you first open the file, but the more you try to navigate, the more broken it becomes. If saving as HTML worked, I'd be satisfied with that, I just need to be able to save a file I can attach to a ticket and send to a colleague.

SamSaffron:
Oh it certainly works, as long as you use `?pp=flamegraph_embed` I just retested.

CyborgMaster:
Flamegraph embed is great for when things are opening in a browser.  This solves the problem of when your server just exposes a JSON API that a single page app hits over AJAX.  In that case, I don't believe there is any other way to get ahold of the flamegraph.  Usually this works by injecting it on page load, but that technique cannot work where there is no page being loaded.  More details are in #350.

OsamaSayegh:
Ok, @CyborgMaster explained his use case in https://github.com/MiniProfiler/rack-mini-profiler/issues/350#issuecomment-656761487 and why `?pp=flamegraph`/`?pp=flamegraph_embed` don't suffice for him. Personally I think this PR is OK, but will defer to @SamSaffron to make the final call.

nateberkopec:
This is now possible using the Speedscope visualizer - just use the export button.

PR #353, opened 2018-05-23
Fix for user agents which does not has Performance.getEntriesByName

yskkin:
In iOS 9.3, window.performance exists but
window.performance.getEntriesByName does not.


SamSaffron:
looks good! thanks!

PR #356, opened 2018-07-13
Add mini-profiler-only route example

bf4:
see https://github.com/MiniProfiler/rack-mini-profiler/blob/1bd8c9fe64940ad514f8f965f7e036c6d8b9d910/README.md#miniprofilers-speed-badge-on-pages-that-are-not-generated-via-rails

per https://github.com/MiniProfiler/rack-mini-profiler/issues/139#issuecomment-192880706


OsamaSayegh:
@bf4 we merged PR https://github.com/MiniProfiler/rack-mini-profiler/pull/396 a while ago that added a special route at `/rack-mini-profiler/requests` which displays MP badge and allow requests profiling in API-only apps, so I don't think we still need to mention how to create a custom route that achieves the same thing? Can we close this?

bf4:
yup!

PR #357, opened 2018-07-16
make _profilin cookie httpOnly and secure if request is over ssl

gregmolnar:


SamSaffron:
are we fussing with the cookie from the client anywhere? Overall I am fine with this change.

gregmolnar:
The reason I propose this change is, when you use miniprofiler in production, it is advised to follow security best practices with the cookies, even though there is nothing sensitive in this cookie (I should've mentioned this at the first comment).

gregmolnar:
@SamSaffron Is there any chance to merge this?

SamSaffron:
I guess we can try this, will merge it now.

PR #358, opened 2018-07-20
Add total sql count

zaru:
Fix to display SQL execution count on button.

I think that it is easier to improve by seeing the execution count of SQL together with the execution time.

Thank you for the awesome gem.


SamSaffron:
Can I see a screenshot?

zaru:
@SamSaffron 
42985914-627d8104-8c2e-11e8-9f77-2d9cf6180302

SamSaffron:
OK so let's make this optional, default off, be sure to add doco to the
readme.

zaru:
@SamSaffron 
Thank you for confirmation.
Changed to option.

```
Rack::MiniProfiler.config.show_total_sql_count = true
```

SamSaffron:
cool I think this is fine!

PR #359, opened 2018-07-26
Amends color of .profiler-unit to pass lighthouse accessibility audit

Otterpocket:
The current style prevents a successful accessibility audit in google lighthouse due to the contrast ratio of the unit (**ms** for example).
Not only does it seem like a good idea to remedy this for usability sake but also to make it less troublesome when running accessibility audits on local projects.

This pr simply uses $textColor (#555) rather than $mutedColor (#aaa) in the css.

### DIFF
--- a/lib/html/includes.css
+++ b/lib/html/includes.css
@@ -35,11 +35,11 @@
     overflow: hidden;
     text-overflow: ellipsis; }
   .profiler-result .profiler-unit {
-    color: #555; }
+    color: #767676; }
   .profiler-result .profiler-trivial {
     display: none; }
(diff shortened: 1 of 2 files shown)


SamSaffron:
hmmm I get that this does not pass lighthouse but feel it is a bit too loud for my liking in black

Why not just add the CSS rule on your side to override it so you can pass lighthouse?

Otterpocket:
See what you think, I find this much easier to read, any lighter doesn't pass accessibility standards.

**This Commit:**
image
image

**Master branch:**
image
image

SamSaffron:
I guess this if fine, looks a bit better after the change.

Issue #354, opened 2018-06-08
MiniProfiler crashes intermittently.


sarojkh:
Lately MiniProfiler has been crashing every now and then, and I can't seem to reproduce the behavior at will. I am not familiar with the intricacies of this gem.

This has been noticed after upgrading to `v1.0.0` (previously when the application had `v0.10.5`, this issue was never observed).

```
Unexpected error while processing request: can't convert nil into Integer
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/timer_struct/page.rb:73:in `%'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/timer_struct/page.rb:73:in `extra_json'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/timer_struct/page.rb:64:in `to_json'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/profiler.rb:104:in `serve_results'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/profiler.rb:129:in `serve_html'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rack-mini-profiler-1.0.0/lib/mini_profiler/profiler.rb:178:in `call'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/railties-5.0.6/lib/rails/engine.rb:522:in `call'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/railties-5.0.6/lib/rails/railtie.rb:193:in `public_send'
	/Users/whatever/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/railties-5.0.6/lib/rails/railtie.rb:193:in `method_missing'
[...]

PR #360, opened 2018-07-30
fix missing started_at on request

gregmolnar:
In #333 the `started_at` attribute it deleted from the `Page` struct(https://github.com/MiniProfiler/rack-mini-profiler/pull/333/files#diff-f8c72372f6f62551dbeeb2ac29be1152R73), and it causes a `nil` issue in some cases. I couldn't really find in the code where this happens(probably `to_json` is called twice on the struct in some cases), but we have a bunch of #354  in production.

### DIFF
--- a/lib/mini_profiler/timer_struct/page.rb
+++ b/lib/mini_profiler/timer_struct/page.rb
@@ -71,7 +71,7 @@ module Rack
 
         def extra_json
           {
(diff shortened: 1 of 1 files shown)


SamSaffron:
yeah interesting, that delete pattern is very not nice.

PR #362, opened 2018-09-12
Reduce memory use of file_store.rb

schneems:
## What

Writing to disk via `File.open` by default is buffered. By disabling the buffer via setting `sync = true` we can reduce memory allocation. It looks like no matter how small the amount of data being written is a default buffer size is used.

## Impact

While making a request to CodeTriage and measuring with memory_profiler via derailed benchmarks I see:

Before:
Total allocated: 742573 bytes (6644 objects)
     48741  rack-mini-profiler/lib

After:
Total allocated: 734236 bytes (6645 objects)
     40477  rack-mini-profiler/lib

This change accounts for 

Diff:

(742573 - 734236) / 742573.0 # => **1.1 % decrease in memory**

## Other Considerations

Is this slower than the buffered option?

```ruby
require 'benchmark/ips'

DATA = ["1ada37891cbb5442fa56157bc451a608", nil, 634602.016795]

def sync
  File.open("tmp-sync.dump", 'w') do |f|
    f.sync = true
    Marshal.dump(DATA, f)
  end
end

def no_sync
  File.open("tmp-no-sync.dump", 'w') do |f|
    Marshal.dump(DATA, f)
  end
end

Benchmark.ips do |x|
  x.report("no-sync") { no_sync }
  x.report("sync   ") { sync }
  x.compare!
end
# Warming up --------------------------------------
#              no-sync   597.000  i/100ms
#              sync      593.000  i/100ms
# Calculating -------------------------------------
#              no-sync      6.404k (±12.2%) i/s -     31.641k in   5.026470s
#              sync         6.416k (±12.4%) i/s -     32.022k in   5.079501s

# Comparison:
#              sync   :     6416.2 i/s
#              no-sync:     6404.3 i/s - same-ish: difference falls within error
```

It looks like the sync is not any slower and in most of my test runs was marginally faster (though well within standard deviation so it's impossible to make that claim in a statistically significant manner).

### DIFF
--- a/lib/mini_profiler/storage/file_store.rb
+++ b/lib/mini_profiler/storage/file_store.rb
@@ -24,7 +24,10 @@ module Rack
         end
 
         def []=(key,val)
-          ::File.open(path(key), "wb+") {|f| f.write Marshal.dump(val)}
+          ::File.open(path(key), "wb+") do |f|
+            f.sync = true
+            f.write Marshal.dump(val)
+          end
         end
 
         private


nateberkopec:
@schneems With benchmark/ips, I usually like to see error numbers (here ±12.2%) less than 5%. What happens if you run the benchmark for longer than the default 5 seconds?

schneems:
Bumping up to 30 seconds still gives me a large error:

```
irb(main):018:0> Benchmark.ips do |x|
irb(main):019:1*   x.report("no-sync") { no_sync }
irb(main):020:1>   x.report("sync   ") { sync }
irb(main):021:1>   x.time = 30
irb(main):022:1>   x.compare!
irb(main):023:1> end
Warming up --------------------------------------
             no-sync   606.000  i/100ms
             sync      592.000  i/100ms
Calculating -------------------------------------
             no-sync      5.999k (±15.5%) i/s -    175.134k in  30.031027s
             sync         6.018k (±16.5%) i/s -    174.640k in  30.074512s

Comparison:
             sync   :     6017.8 i/s
             no-sync:     5999.3 i/s - same-ish: difference falls within error
```

Even with 100 seconds, it's a really high variance. I'm guessing the disk IO is the variance culprit.

schneems:
It was pointed out that  the `file.write` method is faster for lots of data rather than manually passing in a file object to Marshal.dump btw. The PR was already using the `file.write` method so it doesn't need to be updated. Just mentioning for completeness. Thanks to .

```ruby
require 'benchmark/ips'

def rand_str
  (0...50).map { ('a'..'z').to_a[rand(26)] }.join
end
big_data = []
10_000.times { big_data << [rand_str, nil, rand(10_000) + rand() ] }

DATA = big_data

def no_write
  File.open("tmp-sync.dump", 'w+') do |f|
    f.sync = true
    Marshal.dump(DATA, f)
  end
end

def write
  File.open("tmp-no-sync.dump", 'w+') do |f|
    f.sync = true
    f.write Marshal.dump(DATA)
  end
end

Benchmark.ips do |x|
  x.report("no-write") { no_write }
  x.report("write   ") { write }
  x.compare!
end
# Warming up --------------------------------------
#             no-write     4.000  i/100ms
#             write        5.000  i/100ms
# Calculating -------------------------------------
#             no-write     43.150  (± 7.0%) i/s -    216.000  in   5.030772s
#             write        60.228  (± 1.7%) i/s -    305.000  in   5.067139s

# Comparison:
#             write   :       60.2 i/s
#             no-write:       43.1 i/s - 1.40x  slower
```

Fryguy:
Both of these use File.open, where you have this control.  I'm curious; how does File.write stack up?

Fryguy:
https://ruby-doc.org/core-2.4.2/IO.html#method-c-write

Was on mobile at the time I posted, so couldn't test myself... I'll give it a shot in a few hours when I'm back at a computer.

Fryguy:
Tweaked your original script to do "wb+" mode as per the actual PR, and then added all of the modes (file.write, dump direct, and File.write)...

```ruby
require 'benchmark/ips'

DATA = ["1ada37891cbb5442fa56157bc451a608", nil, 634602.016795]

def sync_write
  File.open("tmp-sync.dump", 'wb+') do |f|
    f.sync = true
    f.write Marshal.dump(DATA)
  end
end

def no_sync_write
  File.open("tmp-no-sync.dump", 'wb+') do |f|
    f.write Marshal.dump(DATA)
  end
end

def sync_dump_direct
  File.open("tmp-sync.dump", 'wb+') do |f|
    f.sync = true
    Marshal.dump(DATA, f)
  end
end

def no_sync_dump_direct
  File.open("tmp-no-sync.dump", 'wb+') do |f|
    Marshal.dump(DATA, f)
  end
end

def file_write
  File.write("tmp-no-sync.dump", Marshal.dump(DATA), :mode => "wb+")
end

Benchmark.ips do |x|
  x.report("no-sync write       ") { no_sync_write }
  x.report("sync write          ") { sync_write }
  x.report("no-sync dump direct ") { no_sync_dump_direct }
  x.report("sync dump direct    ") { sync_dump_direct }
  x.report("File.write          ") { file_write }
  x.compare!
end

# Warming up --------------------------------------
# no-sync write          516.000  i/100ms
# sync write             505.000  i/100ms
# no-sync dump direct    495.000  i/100ms
# sync dump direct       493.000  i/100ms
# File.write             579.000  i/100ms
# Calculating -------------------------------------
# no-sync write             5.079k (± 6.8%) i/s -     25.284k in   5.003636s
# sync write                5.252k (± 7.9%) i/s -     26.260k in   5.035655s
# no-sync dump direct       5.006k (± 6.6%) i/s -     25.245k in   5.072092s
# sync dump direct          5.216k (± 6.5%) i/s -     26.129k in   5.032092s
# File.write                5.651k (± 6.6%) i/s -     28.371k in   5.045716s
#
# Comparison:
# File.write          :     5651.3 i/s
# sync write          :     5251.9 i/s - same-ish: difference falls within error
# sync dump direct    :     5216.2 i/s - same-ish: difference falls within error
# no-sync write       :     5078.8 i/s - same-ish: difference falls within error
# no-sync dump direct :     5005.9 i/s - same-ish: difference falls within error
```

Fryguy:
Seems File.write is the "fastest", but the original purpose of this was for memory/allocations, and I'm not sure how you got those numbers.

Separate question, why "wb+" as opposed to "wb"?  The "+" makes it a read/write file handle, but since this is just dumping data, the "+" seems unnecessary.  I'm curious if removing the "+" helps out at all.

schneems:
The "w+" is so if the file doesn't exist it gets created. It makes it easier to share the benchmark.

ohaddahan:
@Fryguy the PR by @schneems slightly changed how to write to the file, which caused a slight degradation (according to the benchmarks).
The `sync = true` memory observation still holds.

Since we only write to the file once, there is no need for a buffer here, even for large data dumps (I benchmarked it and as long as you dump it all in one go, no significant difference based on `sync` value).

I think that `sync write` should be used, I benchmarked it and the results were very close, and in your benchmark they are also very close , can be due to on IO/CPU/OS variance.

But when we run them through memory profiler, `sync write` is superior by far (tried passing `sync: true` to `File.write` seems to ignore it).

`File.write`:

```
Total allocated: 9192 bytes (8 objects)
Total retained:  0 bytes (0 objects)

allocated memory by gem
-----------------------------------
      9192  other

allocated memory by file
-----------------------------------
      9192  debug_file_write_memory.rb

allocated memory by location
-----------------------------------
      9192  debug_file_write_memory.rb:20

allocated memory by class
-----------------------------------
      8424  File
       576  Hash
       120  String
        72  Thread::Mutex
allocated objects by gem
-----------------------------------
         8  other

allocated objects by file
-----------------------------------
         8  debug_file_write_memory.rb

allocated objects by location
-----------------------------------
         8  debug_file_write_memory.rb:20

allocated objects by class
-----------------------------------
         3  Hash
         3  String
         1  File
         1  Thread::Mutex
```

`sync write` : 

```
Total allocated: 352 bytes (4 objects)
Total retained:  0 bytes (0 objects)

allocated memory by gem
-----------------------------------
       352  other

allocated memory by file
-----------------------------------
       352  debug_file_write_memory.rb

allocated memory by location
-----------------------------------
       312  debug_file_write_memory.rb:13
        40  debug_file_write_memory.rb:15

allocated memory by class
-----------------------------------
       232  File
       120  String

allocated objects by gem
-----------------------------------
         4  other

allocated objects by file
-----------------------------------
         4  debug_file_write_memory.rb

allocated objects by location
-----------------------------------
         3  debug_file_write_memory.rb:13
         1  debug_file_write_memory.rb:15

allocated objects by class
-----------------------------------
         3  String
         1  File
```

Fryguy:
> The "w+" is so if the file doesn't exist it gets created. It makes it easier to share the benchmark.

"w" does the same thing: https://ruby-doc.org/core-2.4.2/IO.html#method-c-new-label-IO+Open+Mode

Fryguy:
> tried passing sync: true to File.write seems to ignore it

Yeah, seems you can pass sync via params (File.open can't even take it so you can't even pass via :open_args)

Some of the allocation overhead is probably partly because of the Hash needed to set the mode args.  If the mode is "w" (which is the default for File.write), then there won't be needed, which might save a few allocations.  sync_write may still be more efficient anyway though.

PR #367, opened 2018-12-05
Allow filtering of stack frames in flamegraphs

jstultz:
For some larger Rails applications, the flamegraph can be pretty
overwhelming and full of non-application-related stack frames. While
that information is often valuable, at other times it's really helpful
to be able to see a simpler view of a smaller subset of the entire
process.

This leverages the changes added to flamegraph in
https://github.com/SamSaffron/flamegraph/pull/21 and adds a
rack-mini-profiler configuration to pass through to flamegraph to
achieve the above goal.


jstultz:
(note that for this to actually be usable, there will also need to be a new release of the flamegraph gem, as the above mentioned PR has not yet made it into a released version)

jstultz:
I'm going to close this as it's cluttering up my open PRs and it's been well over a year with no response; I don't actively use this project. If you like the change, take it or leave it.

PR #375, opened 2019-03-20
fix the way that fetch is patched

SampsonCrowley:
You should not be directly overwriting fetch in a way that returns something other than the original fetch. especially since clients might have added additional functionality to the original fetch function


SamSaffron:
nice fix, :+1:

Issue #372, opened 2019-01-30
Sass gem deprecated


P9GIT:
Hi team.

Thanks for this great and useful gem.

This gem depends on the sass gem which is deprecated and shortly will be unmaintained: https://sass-lang.com/ruby-sass

Any plans for migrating away to e.g. the SassC gem?

Best regards, Thomas.

SamSaffron:
I would love to see this migrated to sassc, keep in mind there is no runtime dependency this is only build time so it is not urgent or anything.

P9GIT:
@SamSaffron 

Let's hope so. Sassc should be a drop-in replacement.
Thanks for your answer.

PR #379, opened 2019-04-25
☠️  rip sass ☠️  🎉 long live sassc 🎉 

itsWill:
After a bundle install we get this message: 

```
Ruby Sass has reached end-of-life and should no longer be used.
* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install
* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme
* For more details, please refer to the Sass blog:
  https://sass-lang.com/blog/posts/7828841
```

This replaces it with the recommend `sassc` gem, I had a look at `includes.scss`  and I wasn't able to immediately spot anything that breaks compatibility, the `compile_sass` `rake` task should also keep working without changing anyting. Did a quick test with a `?pp=flamegraph` and all looked good. 

Fixes: https://github.com/MiniProfiler/rack-mini-profiler/issues/372

### DIFF
--- a/rack-mini-profiler.gemspec
+++ b/rack-mini-profiler.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'dalli'
   s.add_development_dependency 'rspec', '~> 3.6.0'
   s.add_development_dependency 'redis'
-  s.add_development_dependency 'sass'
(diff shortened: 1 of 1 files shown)


SamSaffron:
cool! lets try it out!

PR #382, opened 2019-05-22
Test against Ruby 2.6

meganemura:
T/O


SamSaffron:
we got to actually fix the issue here as well :)

PR #383, opened 2019-05-22
Disable Style/FrozenStringLiteralComment cop

meganemura:
#382 is failing due to the rubocop issue.
This fixes it.

By the way, I think the current RuboCop settings are not good for contributors.
Followings are uncontrollable and make build fragile simply.
1. [We don't specify the rubocop version](https://github.com/MiniProfiler/rack-mini-profiler/blob/08ad0c798f40588586c5b841491481fefd847b1c/rack-mini-profiler.gemspec#L37)
2. [.rubocop.yml is inheriting from another repository](https://github.com/MiniProfiler/rack-mini-profiler/blob/08ad0c798f40588586c5b841491481fefd847b1c/.rubocop.yml#L1)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,1 +1,4 @@
 inherit_from: https://raw.githubusercontent.com/discourse/discourse/master/.rubocop.yml
+
+Style/FrozenStringLiteralComment:
+  Enabled: false


meganemura:
I added `frozen_string_literal: true` to every files by `rubocop --auto-correct` at first [(Travis CI builds)](https://travis-ci.org/MiniProfiler/rack-mini-profiler/builds/535875781).
But the changes broke the builds, so I disable Style/FrozenStringLiteralComment cop at the moment.

SamSaffron:
I will see if I can sort this out, thanks!

SamSaffron:
Ok I want to fix this stuff up but we got to keep inheriting from Discourse.

For starter I will follow what I do with all my other projects and drop support for Ruby 2.2 and below as they are EOL, then clean up things internally.

meganemura:
Close in favor of ea182527aec2b43268a1920afd768566abab47ad

PR #386, opened 2019-06-17
Avoid deprecated use of Rack::Request#[]

semaperepelitsa:
When running Ruby in verbose mode ($VERBOSE = true) you can see a warning:
> WARNING Request#[] is deprecated and will be removed in a future version of Rack. Please use request.params[] instead

See https://github.com/rack/rack/pull/1069

"Params" hash must be accessed with string keys, there is no implicit conversion from symbols.


SamSaffron:
Seems perfectly reasonable to me! thanks!

PR #389, opened 2019-06-25
'More' link points to actual request path

markvanlan:
The moreUrl before was set to the windows current location + "?pp=help", which is not as useful when other requests are happening after the initial page load.

sponsored by @nateberkopec's speedshop.

### DIFF
--- a/lib/mini_profiler/timer_struct/request.rb
+++ b/lib/mini_profiler/timer_struct/request.rb
@@ -5,15 +5,15 @@ module Rack
     module TimerStruct
       class Request < TimerStruct::Base
 
-        def self.createRoot(name, page)
-          TimerStruct::Request.new(name, page, nil).tap do |timer|
+        def self.createRoot(name, more_path, page)
+          TimerStruct::Request.new(name, more_path, page, nil).tap do |timer|
             timer[:is_root] = true
           end
         end
 
         attr_accessor :children_duration
 
-        def initialize(name, page, parent)
+        def initialize(name, more_path, page, parent)
           start_millis = (Process.clock_gettime(Process::CLOCK_MONOTONIC) * 1000).to_i - page[:started]
           depth        = parent ? parent.depth + 1 : 0
           super(
@@ -33,6 +33,7 @@ module Rack
             sql_timings_duration_milliseconds: 0,
             is_trivial: false,
             is_root: false,
+            more_path: more_path,
             depth: depth,
             executed_readers: 0,
             executed_scalars: 0,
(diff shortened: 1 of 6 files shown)


SamSaffron:
Thanks! 

The tricky thing here is non GETs, (eg: deletes/puts/patches and so on), maybe only include the "more" link on GET paths then?

markvanlan:
@SamSaffron I didn't think about that. So you are okay if the 'more' link is simply not rendered if request is anything but a GET request?

SamSaffron:
I think just keep it around but point it at root, cause it still can be handy and will be a bit confusing if it part vanishes.

markvanlan:
@SamSaffron Totally. I will put that back in.

lib/mini_profiler/timer_struct/request.rb
```diff
@@ -5,15 +5,15 @@ class MiniProfiler
     module TimerStruct
       class Request < TimerStruct::Base
 
-        def self.createRoot(name, page)
-          TimerStruct::Request.new(name, page, nil).tap do |timer|
+        def self.createRoot(name, more_path, page)
```
SamSaffron: I am so sorry I am pushing this back and forth so often but I do feel we are mixing concerns here.

From the profilers perspective this is simply a concern of `TimerStruct::Page` no? 

It gets the environment, and it can then look up:

- What is the path of this request?
- What is the verb of the request? 

It can then pass those two fields back to the client which in turn can decide what do do (how to create the more path and what to put in it)

markvanlan:
So, since `name` already contains the full path and the request type, you think I should simply parse  `name` in the moreUrl function to create the proper src? 

I hope that was clear enough.

SamSaffron:
Yeah but maybe we also need to ship the verb down to the client?

markvanlan:
Since the verb is always the first word of the name attribute, I can see 2 paths.

1. Change no functionality on the server, and simply parse the verb and path out of the name attribute to create the moreUrl. Basically extending the moreUrl function.

2. Remove the name attribute, to be replaced with verb, and path. Then in the request template construct the name from the 2 attributes. This way there wouldn't be string parsing to create the more link. It would take a bit more work however.

Do either of those sound like the correct direction?

SamSaffron:
I say just go with (1) for now, it is the simplest way forward and amends the least amount of code. 

I like (2) but it is a bigger change.

markvanlan:
@SamSaffron Here is implementation for option 1. Certainly doesn't look as nice as splitting name into the two, but it sure does the trick. Let me know if there is anything else I can improve here.

SamSaffron:
Nice, I am happy with this, thanks so much!

PR #390, opened 2019-07-02
Remove jQuery dependency

majakomel:


SamSaffron:
@markvanlan wondering if you feel like taking over this? We are part done here but still need to work out how to do templates, I was thinking svelte maybe? 

The idea here is to have rack-mini-profile ship a lot less JavaScript and no longer depend on jQuery which is such a big dependency that seems extremely overkill.

markvanlan:
Yeah I will definitely take a look at this! I'll keep you in the loop
before going down a path too far.

SamSaffron:
awesome! thanks so much

mark-vanlandingham-roostify:
@SamSaffron We could use [mustache.js](https://github.com/janl/mustache.js/) to replace the jQuery templating. The behavior would be really similar, but the implementation would be cleaner, and would ship much less js. Do you have any thoughts on that?

markvanlan:
@SamSaffron We could use [mustache.js](https://github.com/janl/mustache.js/) to replace the jQuery templating. The behavior would be really similar, but the implementation would be cleaner, and would ship much less js. Do you have any thoughts on that?

SamSaffron:
Hi Mark, 

I am about to go on a 3 week break, starting on Saturday. This is not a big problem here cause @eviltrout  can still follow up here and provide guidance and merge. 

I was wondering, is svelte an option here? Ideally I want to ship something with no runtime for the templates, mustache is certainly smaller, but can we use it without carrrying a runtime (in a compiled only mode) 

I worry about conflicts and weight here a lot. I want mini profiler to be as minimally intrusive as possible.

markvanlan:
I'll will look into it. Thanks Sam.

 Have a great trip. 3 weeks is amazing. Hope you feel refreshed from it!

SamSaffron:
@markvanlan  I spent some time looking at svelte and yes it is very pretty and I like the compiler approach but there are 2 big blockers here for me: 

1. This would be an almost complete re-write of rack-mini-profiler ui, it would run deep, take ages and most likely regress lots of things. 
2. The bundle size will be bigger from all my internal experiments, the thing does generate a lot of JS

Instead we are going to look at using http://olado.github.io/doT/ to strip out jquery tml. This is a pretty straight forward task. 

Nothing for you specifically to look at now, but thought I should share the news.

SamSaffron:
thanks heaps @majakomel , hope all is going well @OsamaSayegh built on this on #399

PR #391, opened 2019-07-13
Display controller name at ActionController#process

gregmolnar:
It might be just me, but I would find it useful to see the name of the controller, not just the name of the action.

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -58,7 +58,7 @@ module Rack::MiniProfilerRails
 
     # Attach to various Rails methods
     ActiveSupport.on_load(:action_controller) do
-      ::Rack::MiniProfiler.profile_method(ActionController::Base, :process) { |action| "Executing action: #{action}" }
+      ::Rack::MiniProfiler.profile_method(ActionController::Base, :process) { |action| "Executing: #{self.class.to_s}##{action}" }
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am not against doing something here, but I think `self.class.to_s` may not be exactly what we want. 

```
[2] pry(main)> Admin::EmailController.to_s
=> "Admin::EmailController"
```

My guess is that we probably want to display `admin::email#index` or something like that here.

Also ... if we go down this path the other concern is that we just take up too much space so I would change it to:

`admin::email#index` and nuke the `Executing action: ` part to save on space.

gregmolnar:
Removing the `Executing action:` bit would be indeed good, but I think the actual class name is the useful one. This is how it would look with your example by the way:
```
Admin::EmailController#index
```
because the action is still appended: `#{self.class.to_s}##{action}`

SamSaffron:
We could use controller_name. 

```
 pry(main)> Admin::EmailController.new.controller_name
=> "email"
```

We would lose the namespace here which is not ideal but maybe there is another way of grabbing it. 

I find the `admin:email#index` to be the ideal solution here as it is far shorter and the duplication for `Controller` is then removed from a pretty tight space in the UX.

nateberkopec:
Closing as stale but this one is very, very close to being usable if someone wants to pick it up and ship it.

PR #393, opened 2019-07-17
WIP: display activerecord query cache hits

gregmolnar:
Based on @nateberkopec 's recommendation, this change would display the Active Record queries hitting its query cache.

### DIFF
--- a/lib/html/includes.tmpl
+++ b/lib/html/includes.tmpl
@@ -61,6 +61,15 @@
               </td>
             {{/each}}
             </tr>
+            {{if activerecord_query_cache_hits.length > 0}}
+              <tr>
+                <td colspan="3">&nbsp;</td>
+                <td colspan="2" class="profiler-number profiler-percent-in-sql" title="${activerecord_query_cache_hits} queries served by ActiveRecord::QueryCache">
+                  <a class="cached-queries-show">${activerecord_query_cache_hits.length}</a>
+                  <span class="profiler-unit"> queries served by ActiveRecord::QueryCache</span>
+                </td>
+              </tr>
+            {{/if}}
           </tfoot>
         </table>
         {{if client_timings}}
@@ -121,7 +130,23 @@
       </p>
     </div>
   {{/if}}
-
+  {{if activerecord_query_cache_hits.length > 0}}
+    <div class="cached-queries">
+      <table>
+      <thead>
+        <tr>
+          <th style="text-align:left">Query</th>
+          <th style="text-align:left">Location</th>
+        </tr>
+      </thead>
+      <tbody>
+        {{each(i, s) activerecord_query_cache_hits}}
+          {{tmpl({g: s}) "#queryCacheTemplate"}}
+        {{/each}}
+      </tbody>
+      </table>
+    </div>
+  {{/if}}
   </div>
 
 </script>
@@ -225,4 +250,15 @@
     </td>
   </tr>
 
+</script>
+
+<script id="queryCacheTemplate" type="text/x-jquery-tmpl">
+  <tr>
+    <td class="query">
+      <div>${g.sql}</div>
+    </td>
+    <td class="query">
+      <div>${g.location}</div>
+    </td>
+  </tr>
 </script>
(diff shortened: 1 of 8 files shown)


lib/html/includes.tmpl
```diff
@@ -61,6 +61,15 @@
               </td>
             {{/each}}
             </tr>
+            {{if activerecord_query_cache_hits.length > 0}}
+              <tr>
+                <td colspan="3">&nbsp;</td>
+                <td colspan="2" class="profiler-number profiler-percent-in-sql" title="${activerecord_query_cache_hits} queries hit the cache">
```
nateberkopec: my preferred wording here would be "served by ActiveRecord::QueryCache", idk if that fits/is too lengthy.

lib/mini_profiler/profiler.rb
```diff
@@ -76,6 +76,17 @@ def initialize(app, config = nil)
         @config.storage_instance = @config.storage.new(@config.storage_options)
       end
       @storage = @config.storage_instance
+      if defined?(ActiveSupport)
+        ActiveSupport::Notifications.subscribe "sql.active_record" do |*args|
+          event = ActiveSupport::Notifications::Event.new *args
+          if event.payload[:cached]
```
nateberkopec: the fifth argument to subscribe is the hash payload, no need to create a new event object here I think?

lib/mini_profiler/profiler.rb
```diff
@@ -76,6 +76,17 @@ def initialize(app, config = nil)
...
+            current.current_timer.page.attributes[:activerecord_query_cache_hits] << {
+              sql: event.payload[:sql],
+              binds: event.payload[:binds].collect { |b| "#{b.name}: #{b.value_for_database}" }.join(', ')
```
nateberkopec: I think you meant for this to be a map?

lib/mini_profiler/profiler.rb
```diff
@@ -76,6 +76,35 @@ def initialize(app, config = nil)
         @config.storage_instance = @config.storage.new(@config.storage_options)
       end
       @storage = @config.storage_instance
+      if defined?(ActiveSupport)
+        ActiveSupport::Notifications.subscribe "sql.active_record" do |*args|
```
SamSaffron: sadly this is far from a zero cost thing @nateberkopec , in fact at Discourse we do stuff like this:

https://github.com/discourse/discourse/blob/b5705348b32fff406d8e7e3463def65eaf635206/config/initializers/300-perf.rb#L3-L12

I am ok with adding this stuff to rack-mini-profiler, but we are going to new a setting here and to default it off explaining that this adds cost if you turn it on. 

Ideally we should measure what the actual cost of this extra subscriber is.

nateberkopec: Interesting!

I wonder if there's a better way to get this stuff, because all we care about is the caller locations when QueryCache is hit. We don't actually have to subscribe to _all_ AR notifications.

nateberkopec: > I am ok with adding this stuff to rack-mini-profiler, but we are going to new a setting here and to default it off explaining that this adds cost if you turn it on.

I mean this is fine with me too tbh

gregmolnar: I will add a configuration option and default it to be off.

nateberkopec: I think ideally it would be in pp=help and work the same way no-backtrace, normal-backtrace and full-backtrace does. Maybe call it no-ar-cache (default), show-ar-cache (option).

That way you could still use it in production and just flip it back off when you're done. You can unsubscribe from ActiveSupport notifications, right?

SamSaffron:
Yes you can unsubscribe , sure that sounds like a great plan !

guiracine:
Hi! I see that this PR is 3 years old, should it be closed ? I just started to use the profiler and I was looking for a way to distinguish queries that have hit or no hit on the cache.

prostko:
Ship it! :)

nateberkopec:
Closing as stale but this is a good idea so I'm moving it to an issue.

PR #394, opened 2019-07-24
Replace jQuery with compiled Handlebars templates

markvanlan:
@SamSaffron What do you think about this approach? Using pre-compiled handlebars templates?

The huge pain is moving all the logic out of the templates, and having the data be formatted, when it is passed into the templates.


SamSaffron:
I am somewhat concerned here that we would be replacing one devil with another devil. 

The big drivers for removing jQuery is to drive away the dependency so: 

1. We never conflict with the jQuery version the user decided to run locally 
2. We don't ship a massive jQuery dependency
3. We get rid of jQuery tmpl which is largely abandonware at the moment. 

I worry that with handlebars we are now risking a similar issue where we are stuffing info into the global context and could have multiple handlebar runtimes fighting over which version is golden.

Is there any way we can use some templating framework that compiles down to straight JS with no dependencies? Does svelte do this? 

Keep in mind we can be fairly aggressive with the browser target here cause people using mini profiler are expected to be using modern browsers.

SamSaffron:
closing this @markvanlan , @OsamaSayegh is pretty much done now moving us to DoT

PR #395, opened 2019-08-26
Removes use of eval() as JSON is very well supported now

jjaffeux:
We can now safely rely on JSON being available:

https://caniuse.com/#feat=json

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -486,8 +486,7 @@ var MiniProfiler = (function () {
                 // should be an array of strings, e.g. ["008c4813-9bd7-443d-9376-9441ec4d6a8c","16ff377b-8b9c-4c20-a7b5-97cd9fa7eea7"]
                 var stringIds = xhr.getResponseHeader('X-MiniProfiler-Ids');
                 if (stringIds) {
-                    var ids = typeof JSON != 'undefined' ? JSON.parse(stringIds) : eval(stringIds);
-                    fetchResults(ids);
+                    fetchResults(JSON.parse(stringIds));
                 }
             }
         };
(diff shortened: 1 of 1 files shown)


SamSaffron:
hmmm everything about this is strange... the duplication.. the json... why even do json here to get a bunch of ids we should simply split on `,` and be done?

SamSaffron:
@OsamaSayegh can you have a look at ^^^^

jjaffeux:
> @OsamaSayegh can you have a look at ^^^^

Sorry thought I had answered this...

So anyways I didn't want to refactor any deep design of rack-mini-profiler with this commit, but essentially, this is coming from:

https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler/profiler.rb#L383

Which is basically serializing an array of ids using `JSON.generate` https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler/profiler.rb#L589

So we get somethings like : 

```
X-Miniprofiler-Ids: ["4spklr7d0rii7l9ylmoh","t00hjk19bvuplp9l2fgi","f4hp116ac05db1z5ospi","xzim6sud1g2nhvxqx4hi","nv9q4578gqbs0xpq8juh"]
```
And then use `JSON.parse` on the client to deserialize it, so we could do a ids.join(',') on the ruby side and a split(",") on the client side, that would work too, I'm unsure on the reasons if this decisions at the beginning though. It seems like there was a reason, or why go through the specific IE case otherwise?

SamSaffron:
I say we change the header, it seems a bit pointless to do all this work.

coding-horror:
Should @OsamaSayegh  be looking at this? I don't like that it is stalled

SamSaffron:
Sure, once Osama is done with keyboard shortcuts he can sort this one out.

OsamaSayegh:
I can't push to this PR, so I've included this change in https://github.com/MiniProfiler/rack-mini-profiler/pull/399. Thanks @jjaffeux

jjaffeux:
Cool, closing this then! thanks

Issue #243, opened 2016-05-17
Mini Profiler UI not displayed when the response is JSON


cireficc:
In a Rails JSON API, the profiler UI is not displayed, which is unfortunate!

[A workaround is easy though](http://stackoverflow.com/a/26801183/1986871); you need two tabs open - one for the JSON, one for the profiler. The second tab needs to be any page that renders HTML (for example the root route `/`). Make the JSON request, then refresh the page that renders HTML, and the profiler UI will show the JSON request.

Can this be fixed so that the profiler UI shows up on JSON-only pages, or is this by design (and why)?

seanlinsley:
It's impossible for the profiler to show up on JSON requests, because the browser only expects JSON. For HTML requests, we insert JS to render the profiler.

The only way to get around that is by making rack-mini-profiler a browser extension, but that's a huge amount of work.

kbrock:
The request is to actively push events to the browsers.

The infrastructure is not in place for this. It seems websockets /server sent events is overkill here.

Maybe when chrome adds adds header cookies for performance data this may become more viable

SamSaffron:
A custom page that does a req say every 5-10 seconds to update the list should do the trick without enormous amounts of added complexity. Alternatively you could use message bus on such a page. 

Totally support shipping a sample or even a custom page for dealing with this.

southwolf:
The profiler UI will not show up on /404.html in production environment for a rails-api project.

The `Rack::MiniProfiler.authorize_request` configuration won't help

Repo: https://github.com/southwolf/rack-profiler-in-production

kbrock:
Thanks @southwolf could you create a PR for your issue? This seems like a separate thread.

oyeanuj:
@SamSaffron or others around, what is the recommended way of using rack-mini-profiler with Rails in API only mode?

benterprise:
Also have an API only app that I'd like to monitor

printercu:
Maybe it's possible to add new `?pp=` flag which will suppress application output and render profiler layout for that request?

SamSaffron:
@oyeanuj / @printercu  the workaround is in the OP use two browser windows, one open to an HTML page. 

I think we should ship a blank page people can just hit so you can do `http://myapp.com/rack-mini-profiler/requests` or something to see everything. Then you don't have to roll out your own blank page.

RudyOnRails:
[...]

PR #396, opened 2019-08-28
Add blank HTML page at /rack-mini-profiler/requests

mculp:
This PR is to address https://github.com/MiniProfiler/rack-mini-profiler/issues/243


SamSaffron:
Yeah this is a nice quality of life change for api only apps! thanks.

PR #399, opened 2019-09-16
Remove jQuery

OsamaSayegh:
This PR removes jQuery as a dependency and replaces jQuery templating library with a lightweight library called [doT.js](http://olado.github.io/doT/).

Notes:

- We can live without shipping the doT.js source code by pre-compiling the templates into raw JS functions (in a rake task as a part of the `build` task) and ship those instead. But the library is very small (3.34 KB as of version 1.1.2) and the size of raw functions is 12 KB and that can grow as the templates grow.

- I noticed that the `list.js` (and the `list.tmpl` etc.) files weren't referenced anywhere and not being used and the last commit that changes those files was from 6 years ago, so I removed them as they don't seem to be used any more. Please correct me if I'm wrong.

- The `compile_sass` rake was no longer working since the `sass` gem was replaced with `sassc` which doesn't seem to ship with an executable that we can use, so I changed the rake task to use `SassC::Engine` directly.

### Credits

Most of the work here was done by Maja Komel, thank you Maja!

cc @SamSaffron


SamSaffron:
wonderful!!! Thanks heaps!

SamSaffron:
Do we need to pull in mousetrap here? can we just do the keyboard binding by hand?

OsamaSayegh:
Yes we can. The toggle shortcut is configurable via `toggle_shortcut`, so I'll need to write something that can handle different key combinations. I wonder how fancy this should be because I'm reading through jQuery hotkeys https://github.com/jeresig/jquery.hotkeys/blob/master/jquery.hotkeys.js source code and there seem to be a few edge cases?

SamSaffron:
oh I see, I guess just do something minimal here. as long as we support the obvious stuff people can submit Prs for enhancements, I bet 99.999% of installs don't override.

OsamaSayegh:
Ok, I've remove Mousetrap and built something very simple to support different shortcuts (I took inspiration from jQuery hotkeys). Let me know what you think!

SamSaffron:
OK ... I am merging this !!!! thanks so much @OsamaSayegh ... also giving you access to the repo for followup fixes!

PR #402, opened 2019-09-30
Disable risky memory debugging tools by default

OsamaSayegh:
This adds a new config option `enable_advanced_debugging_tools` which is false by default. These query params `?pp={profile-gc,profile-memory,analyze-memory,env}` will only work if the new option is set true, otherwise you'll get a message saying it's disabled and how to enable it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,6 @@ toggle_shortcut|Alt+P|Keyboard shortcut to toggle the mini_profiler's visibility
 start_hidden|`false`|`false` to make mini_profiler visible on page load.
 backtrace_threshold_ms|`0`|Minimum SQL query elapsed time before a backtrace is recorded.
 flamegraph_sample_rate|`0.5`|How often to capture stack traces for flamegraphs in milliseconds.
-disable_env_dump|`false`|`true` disables `?pp=env`, which prevents sending ENV vars over HTTP.
 base_url_path|`'/mini-profiler-resources/'`|Path for assets; added as a prefix when naming assets and sought when responding to requests.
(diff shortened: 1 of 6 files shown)


lib/mini_profiler/profiler.rb
```diff
@@ -308,11 +323,13 @@ def call(env)
       end
 
       if query_string =~ /pp=env/ && !config.disable_env_dump
```
OsamaSayegh: @SamSaffron there is an existing config option for env, should I deprecate this one in favor of the new option?

SamSaffron: yes lets remove that in favor of the new one! 

Once done feel free to merge.

README.md
```diff
@@ -333,6 +333,7 @@ collapse_results|`true`|If multiple timing results exist in a single page, colla
 max_traces_to_show|20|Maximum number of mini profiler timing blocks to show on one page
 html_container|`body`|The HTML container (as a jQuery selector) to inject the mini_profiler UI into
 show_total_sql_count|`false`|Displays the total number of SQL executions.
+enable_advanced_debugging_tools|`false`|Enables memory debugging tools that can be used via the UI.
```
SamSaffron: A bit more of an explanation of why? Also railstie should enable this in development mode.

"Enables sensitive debugging tools that can be used via the UI. In production we recommend keeping this disabled as memory and environment debugging tools can expose contents of memory that may contain passwords."

OsamaSayegh:
Ok, thank you for the review! I've made the changes and merging this now 🚀

PR #403, opened 2019-10-03
Fix CI build failure

meganemura:
To avoid Layout/SpaceAroundOperators offence, added blank lines and spaces.


SamSaffron:
thanks!

PR #405, opened 2019-10-08
Prepare parametrized PG query for showing query parameters in web panel

unavailabl3:
Since Rails 5 has parametrizied query statement like this:
`SELECT * FROM 'items' WHERE 'item'.'id' = $1 [['id',123]]`
So in panel we see `$1` instead of `123`
Params are not shown in rack-mini-profiler panel. This commit is a hack (ONLY FOR POSTGRESQL) to see params. Like this we can add for other DBs.
P.S. May be commit needs refactoring for optimization

### DIFF
--- a/lib/patches/db/pg.rb
+++ b/lib/patches/db/pg.rb
@@ -39,6 +39,20 @@ class PG::Connection
     alias_method :exec_params_without_profiling, :exec_params
   end
 
+  def prepare_parameterized_statement(args)
+    if Array === args[1] && args[1].length > 0
+      query = args[0]
+      parameters = args[1]
+      counter = 0
+      loop do
+        break if !query.include? "$#{counter+=1}"
+        parameter = parameters[counter-1].is_a?(String) ? "'#{parameters[counter-1].gsub("'","''")}'" : parameters[counter-1]
+        query = query.sub("$#{counter}",parameter.to_s)
+      end
+    end
+    query
+  end
+
   def prepare(*args, &blk)
     # we have no choice but to do this here,
     # if we do the check for profiling first, our cache may miss critical stuff
@@ -58,7 +72,8 @@ class PG::Connection
     start        = Process.clock_gettime(Process::CLOCK_MONOTONIC)
     result       = exec_without_profiling(*args, &blk)
     elapsed_time = SqlPatches.elapsed_time(start)
-    record       = ::Rack::MiniProfiler.record_sql(args[0], elapsed_time)
+    query_with_params = prepare_parameterized_statement(args)
+    record       = ::Rack::MiniProfiler.record_sql(query_with_params, elapsed_time)
     result.instance_variable_set("@miniprofiler_sql_id", record) if result
 
     result


lib/patches/db/pg.rb
```diff
@@ -39,6 +39,20 @@ class PG::Connection
     alias_method :exec_params_without_profiling, :exec_params
   end
 
+  def prepare_parameterized_statement(args)
+    if args[0].match(/\$[1-9]/).present? && args[1].is_a?(Array)
```
SamSaffron: .match can be expensive here on every sql query, maybe check for the array first? is there any need to run a regex if you have an array and the array has a length > 0 

```
if Array === args[1] && args[1].length > 0
```

unavailabl3: Refreshed commit with your advice
[](https://github.com/MiniProfiler/rack-mini-profiler/pull/405/commits/584553ea27f72b80572e00962c0130d714e8ae3a)

lib/patches/db/pg.rb
```diff
@@ -58,6 +71,7 @@ def exec(*args, &blk)
     start        = Process.clock_gettime(Process::CLOCK_MONOTONIC)
     result       = exec_without_profiling(*args, &blk)
     elapsed_time = SqlPatches.elapsed_time(start)
+    args[0] = prepare_parameterized_statement(args)
```
SamSaffron: its confusing this is rewriting args[0 just give it a new var imo

lib/patches/db/pg.rb
```diff
@@ -39,6 +39,19 @@ class PG::Connection
...
   end
 
+  def prepare_parameterized_statement(args)
+    if Array === args[1] && args[1].length > 0
+      query = args[0]
+      parameters = args[1]
+      counter = 0
+      loop do
+        break if !query.include? "$#{counter+=1}"
+        query = query.sub("$#{counter}",parameters[counter-1].to_s)
```
SamSaffron: subbing is a bit confusing and can cause a nonsense query to be logged in some cases. For example: 

$1 -> hello world's, which technically would need to be `'hello world''s'` I think

SamSaffron: an alternative may be a preamble comment that explains the param so there is not confusion.  also I think you technically want gsub here, also what if more than 10 params? 

`$1` -> `\* $1 *\ 'hello-world''s'`

unavailabl3: What will happen if more than 10 params?

nateberkopec:
This is still an issue on `master`, and there are no conflicts here, so despite the age I will leave this open.

kbrock:
I like having the parameters separate. I want to know what is actually sent to the server. (easier to detect N+1).
But I use my own rack-mini-profiler client to fetch the sql, so I probably don't see the same issue.

Is the issue that the parameters are not showing in the UI?
Seems like a js/ui issue not a storage issue.

But comment if you feel we should modify the sql before we store it. I can look at this.

Also, historically, I was under the impression that people were concerned about the parameter values being leaked. It may be the reason the parameters were not included into the ui. Not sure if the hide parameter regex met people's concerns.

kbrock:
Then again, I always put the params back into the query when running it.
Could we do this when sending down the queries to the ui?

PR #407, opened 2019-10-13
Replace stale recipe for custom middleware ordering with description of current behaviour

marselmustafin:
Recipe for using `Rack::Deflate` is not suitable for use with Rails 5+: #242
There was an update [#258](https://github.com/MiniProfiler/rack-mini-profiler/pull/258) with disabling of compression when mini-profiler is used with `Rack::Deflate`, so I think it would be better to replace that stale recipe with description of this behaviour


OsamaSayegh:
Yes I think this makes sense. Thanks @marselmustafin!

PR #409, opened 2019-10-22
FIX: string interpolation was breaking ie11 support

jjaffeux:


jjaffeux:
WIP!

I asl need to remove fat arrow and spread

SamSaffron:
should we just run our current include.js through babel and check that in (with minor touchups) ?

jjaffeux:
This is what I did here, manually.

OsamaSayegh:
Ok I've made some changes and it should be now 100% compatible with IE11 with minimal costs.

Notable changes:

* I replaced the spread operator `...` with a simple function `toArray` that loops through `NodeList` using a for loop and returns a normal array. That way we don't need to polyfill `Symbol`.

* Added a small polyfill for `Element.closest` (and `Element.matches` which is needed for `closest` polyfill).

* `forEach` isn't available for `NodeList` in IE11, so I wrapped all `querySelectorAll` calls that are chained to `forEach` with the `toArray` function that I wrote to replace spread operators.

* `Element.remove` isn't available in IE11, so I replaced it with `element.parentElement.removeChild(element)`.

Can you review @jjaffeux @SamSaffron please?

SamSaffron:
> Added a small polyfill for Element.closest (and Element.matches which is needed for closest polyfill).

I would prefer if the polyfill was done a bit differently, don't touch Element.closest, instead use a helper function to do it for now. I worry that polyfills fight with other polyfills so I would prefer to be isolated.

SamSaffron:
overall this is looking great to me, be sure to also pass it through prettier.

SamSaffron:
cool I am merging this in and cutting a release!

PR #411, opened 2019-10-23
Get rid of ProfilingMethods#singleton_class

casperisfine:
It was introduced in https://github.com/MiniProfiler/rack-mini-profiler/pull/190, probably with pre MRI 1.8.x support in mind.

But now rack-mini-profiler requires 2.3+ so it's no longer needed.

If it was named differently it wouldn't matter, but right now it's overriding and changing the visibility of `Object#singleton_class`.


SamSaffron:
oh ouch, yeah makes sense. thanks!

PR #412, opened 2019-10-25
Fix ISO 8601-notation of dates in README

Timmitry:


SamSaffron:
oops had one more, cleaning this up.

PR #413, opened 2019-10-25
Precompile doT.js templates

OsamaSayegh:
This PR adds a build step that pre-compiles our doT.js templates. This is needed because doT.js uses `eval` which prevents us from disallowing `unsafe-eval` in CSP.


SamSaffron:
Awesome change ... going to push out another release now!

PR #416, opened 2019-11-03
Standardize CHANGELOG headings

### LINKED ISSUE #347: Typo in CHANGELOG.md
Hi, I think the year for release 1.0.0 is wrong. It says 2017, it should be 2018.

jonathanhefner:
The primary purpose of this commit is to fix the release date typo for v1.0.0 ("2017-03-29" => "2018-03-29"), but we might as well clean up the other headings too.

Fixes #347.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,20 @@
 # CHANGELOG
 
-## 1.1.3 2019-10-28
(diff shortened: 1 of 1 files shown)


SamSaffron:
Thank you so much @jonathanhefner  for helping out here and with all the old issues! very much appreciated.

PR #417, opened 2019-11-11
Update Redis example configuration

alassek:
If you're using TLS to secure your Redis connection, the given example
will not work correctly because it will choose the incorrect protocol.
You can pass a URL directly to Redis instead of parsing it yourself,
and this gives you the ability to switch between plaintext and TLS with
a single config update.


SamSaffron:
Looks good, simpler is better!

PR #418, opened 2019-11-13
Allow Mini Profiler patches to be optional in Rails applications

OsamaSayegh:
This PR allows Rails application to use mini profiler without any of the monkey patches that mini profiler ships with, instead mini profiler will rely completely on [`ActiveSupport::Notifications`](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#subscribing-to-an-event). 

To use the non-patches version in your Rails app, add `config.mini_profiler_without_patches = true` to your environment configurations file(s).

There will be some differences between the patches and no-patches versions, for example most importantly measuring SQL/actions/render time won't be as accurate, and there won't be a `Net::HTTP` "step" because there is no official API that we can use to replicate what this patch does https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/patches/net_patches.rb.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,11 +16,6 @@ Style/AndOr:
 Style/FrozenStringLiteralComment:
   Enabled: true
 
-# Do not use braces for hash literals when they are the last argument of a
-# method call.
-Style/BracesAroundHashParameters:
-  Enabled: true
-
 # Align `when` with `case`.
 Layout/CaseIndentation:
   Enabled: true
--- a/Rakefile
+++ b/Rakefile
@@ -14,7 +14,9 @@ task default: [:rubocop, :spec]
 require 'rspec/core'
 require 'rspec/core/rake_task'
 RSpec::Core::RakeTask.new(:spec) do |spec|
-  spec.pattern = FileList['spec/**/*_spec.rb']
+  pattern = ARGV[1] || 'spec/**/*_spec.rb'
+  spec.pattern = FileList[pattern]
+  spec.verbose = false
 end
 
 desc "builds a gem"
--- a/lib/disable_method_patches.rb
+++ b/lib/disable_method_patches.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Rack
+  class MiniProfiler
+    DISABLE_METHOD_PATCHES = true
+  end
+end
--- a/lib/enable_rails_patches.rb
+++ b/lib/enable_rails_patches.rb
@@ -2,6 +2,6 @@
 
 module Rack
   class MiniProfiler
-    DISABLE_METHOD_PATCHES = true
+    ENABLE_RAILS_PATCHES = true
   end
 end
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -6,6 +6,10 @@ module Rack
 
       include Rack::MiniProfiler::ProfilingMethods
 
+      def disable_method_patches?
+        !!defined?(::Rack::MiniProfiler::DISABLE_METHOD_PATCHES)
+      end
+
       def generate_id
         rand(36**20).to_s(36)
       end
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -67,6 +67,17 @@ module Rack
           This feature is disabled by default, to enable set the enable_advanced_debugging_tools option to true in Mini Profiler config.
         TEXT
       end
+
+      def binds_to_params(binds)
+        return if binds.nil? || config.max_sql_param_length == 0
+        # map ActiveRecord::Relation::QueryAttribute to [name, value]
+        params = binds.map { |c| c.kind_of?(Array) ? [c.first, c.last] : [c.name, c.value] }
+        if (skip = config.skip_sql_param_names)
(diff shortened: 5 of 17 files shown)


SamSaffron:
I think we should be a bit more official about this support including changes to the README.md

Our default I think is correct, we much prefer to get rich information vs limited. But we should be transparent about it.

After `enable_advanced_debugging_tools` 

`enable_method_patching` : `Allow mini profiler to patch existing classes with instrumentation to gather accurate statistics, if omitted we will attempt to use library specific hooks which existing in ActiveRecord.`


Something like that... Ideally we can also support Sequel here using existing Sequel hooks, perhaps @jeremyevans can help point you at what hooks you can lean on there?

OsamaSayegh:
Sure, I can add a note about this change in README.md and changelog, however I wanted to point out (and make sure you noticed) that I'm not using the `Rack::MiniProfiler.config` object to turn on/off the patches; I'm relying on `Rails.configuration` to turn patches on/off for this reason:

The `Rack::MiniProfiler.config` object is only available after you do require rack-mini-profiler, and requiring mini profiler loads everything including the patches (because the default is true), so running `Rack::MiniProfiler.config.enable_method_patching = false` after require would be way too late. As far as I'm aware there is no clean way to reverse/unload patches in ruby? Maybe if we don't want to tie this config option to Rails, we can use an ENV variable to turn patches on/off?

SamSaffron:
Tricky business! 

One possible idea is to codify this into the `require` used in the Gemfile.

```
gem :rack_mini_profiler, require: [ 'rack_mini_profiler/disable_method_patches', 'rack_mini_profiler' ]
```

The nice thing about this is that it is super clear what is going on and how to move to the "default" way it works if you don't like the money patches.

I am kind of liking this cause it is super explicit about what is going on, and works with straight rack or rails and can easily be documented. 

@rafaelfranca what are your thoughts?

OsamaSayegh:
I really like this idea as well (didn't know `require:` option could take an array!) so I made this change.

If people want to disable patches and want to have `require: false` in their gemfile, they can do this:

```ruby
gem 'rack-mini-profiler', require: ['disable_method_patches']
```

If they want to disable patches but don't want to `require: false` mini profiler, they can do this:

```ruby
gem 'rack-mini-profiler', require: ['disable_method_patches', 'rack-mini-profiler']
```

And no change is needed for existing users.

rafaelfranca:
Having this on gemfile of Rails application is a no-go for me. It is aesthetically  unpleasant and still give the message that patching Rails is ok. Even adding a config to disable monkey patch Rails is a no-go to me. It needs to not patch Rails at all by default to be in the Rails Gemfile.

BTW, I don't have a problem with matching Net::HTTP, I just don't want to add by default a gem to Rails applications that can break every time we change private API that are being patched by the gem.

SamSaffron:
Understood, I am fine to amend the gem not to patch rails by default, it is
the other patches that I need on by default

We can work out the rails missing APIs.

The performance and richness of direct patches to db drivers is not
something I want to give up, I don’t mind working with pg and MySQL long
term to get hooks direct in the db drivers so we can avoid this, but short
term this is tricky

I think the only issue is the action view patch? What changes do we need to
make in rails to avoid the need for it?

rafaelfranca:
Where are the action view patches? I only saw commented code related to Action View.

SamSaffron:
I think this is the sticking point: 

https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler_rails/railtie.rb#L60-L65

I am glad to redo this thing so our fallback is less rich if we must https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/patches/db/activerecord.rb and then by default we have "direct db driver" patches and a fallback of a limited notification.

rafaelfranca:
I think both of those methods send notifications through `ActiveSupport::Notifications`.

SamSaffron:
The trouble with the notifications API is that we are building a tree of events, we ideally want a "before" and "after" call so we can cleanly construct a tree. AS notifications only gives us an after call. We can reverse it into the tree with a bit more work, provided we are lucky and times line up.

I did want to spend a bit of time writing a proposal for a more efficient AS notifications API that can handle these use cases better, will hack away at something.  

But yeah current API should in theory be somewhat workable, albeit less efficient than the method patching.

OsamaSayegh:
Ok, sorry for the delay here 😅 

I've changed the default here so that Rails is not patched at all, but the DB drivers (e.g. `PG`) and `Net::HTTP` are still patched by default. People can opt in Rails patching by requiring the `enable_rails_patches` file in their `Gemfile`.

Note I've used `ActiveSupport::Notifications.monotonic_subscribe` here which is fairly new API in ActiveSupport and hasn't been included in a release yet as far as I know, so this will only work with future versions of Rails.

I've tested this with a brand new Rails app based off Rails master and it worked great (will double test this tomorrow because I've made some small changes while I was writing tests).

SamSaffron:
@rafaelfranca thoughts? is this in line with what you were thinking?

rafaelfranca:
Yes. Great work on this! I think you will have to deal with `monotonic_subscribe` given it is not released yet. Maybe check of its existence before calling?

OsamaSayegh:
Thank you @rafaelfranca! I've updated the PR so that Mini Profiler will fallback to `subscribe` if `monotonic_subscribe` is not available.

PR #419, opened 2019-11-16
Fix grammar in README

alberto:


SamSaffron:
My Yoda Speak only takes me ... that ... far

:heart:  any improvements to the readme are super welcome.

PR #425, opened 2019-12-03
using be_within matcher to avoid float rounding issues by calling to_i

### LINKED ISSUE #422: Flaky specs in spec/lib/profiler_spec.rb
While opening a PR for the project I noticed due to one failng build that there are flaky specs:

```
rspec ./spec/lib/profiler_spec.rb:116 # Rack::MiniProfiler step typical usage measures total duration correctly

rspec ./spec/lib/profiler_spec.rb:124 # Rack::MiniProfiler step typical usage measures outer duration correctly
```

The related build: https://travis-ci.org/MiniProfiler/rack-mini-profiler/jobs/617729576?utm_medium=notification

While renunning those tests they were green again. So there seems to be a timing issue because the expected values are off only by one second. 

Bildschirmfoto 2019-11-27 um 14 52 46

jethroo:
This PR fixes #422 which was happening due to an rounding error by converting float into int with `to_i`. Rather than converting know the `be_within` matcher is used for the expectations.

### DIFF
--- a/spec/lib/profiler_spec.rb
+++ b/spec/lib/profiler_spec.rb
@@ -101,6 +101,7 @@ describe Rack::MiniProfiler do
           }
           clock_set(start + 1 + 2 + 3 + 4)
         }
+
         @page_struct = Rack::MiniProfiler.current.page_struct
         @root = @page_struct.root
         @root.record_time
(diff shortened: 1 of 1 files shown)


SamSaffron:
Awesome, thanks for this fix!

PR #426, opened 2019-12-12
SECURITY: prevent XSS from malicious SQL query

lis2:
The carefully prepared SQL query can be evaluated by javascript.

To prevent it, we need to escape HTML from the query.


SamSaffron:
thanks heaps... getting another gem out.

Issue #406, opened 2019-10-10
mini profiler version 1.1.0 gives various errors in browser console and doesn't seem to work properly


kikonen:
It seems that mini profiler is no (with version 1.1.0) dumping various internal errors into console, and profiling of XHR requests doesn't seem work any longer.

Errors dumped in startup
```

Refused to get unsafe header "X-MiniProfiler-Ids"

VM654:3 Uncaught ReferenceError: duration_milliseconds is not defined
    at Object.eval [as profilerTemplate] (eval at r.template (dot.1.1.2.min.js?v=0ade3000608cde95db93022b6576d63a:2), <anonymous>:3:2622)
    at renderTemplate (includes.js?v=0ade3000608cde95db93022b6576d63a:187)
    at buttonShow (includes.js?v=0ade3000608cde95db93022b6576d63a:194)
    at XMLHttpRequest.request.onload (includes.js?v=0ade3000608cde95db93022b6576d63a:158)

VM654:3 Uncaught ReferenceError: duration_milliseconds is not defined
    at Object.eval [as profilerTemplate] (eval at r.template (dot.1.1.2.min.js?v=0ade3000608cde95db93022b6576d63a:2), <anonymous>:3:2622)
    at renderTemplate (includes.js?v=0ade3000608cde95db93022b6576d63a:187)
    at buttonShow (includes.js?v=0ade3000608cde95db93022b6576d63a:194)
    at XMLHttpRequest.request.onload (includes.js?v=0ade3000608cde95db93022b6576d63a:158)

VM654:3 Uncaught ReferenceError: duration_milliseconds is not defined
    at Object.eval [as profilerTemplate] (eval at r.template (dot.1.1.2.min.js?v=0ade3000608cde95db93022b6576d63a:2), <anonymous>:3:2622)
    at renderTemplate (includes.js?v=0ade3000608cde95db93022b6576d63a:187)
    at buttonShow (includes.js?v=0ade3000608cde95db93022b6576d63a:194)
    at XMLHttpRequest.request.onload (includes.js?v=0ade3000608cde95db93022b6576d63a:158)

Refused to get unsafe header "X-MiniProfiler-Ids"
```

boo68120:
Did u got the solution for  
Refused to get unsafe header "X-MiniProfiler-Ids"

kikonen:
My solution so far was to revert back to older version of rack-mini-profiler.

SamSaffron:
[...]

PR #428, opened 2020-01-04
FIX: Custom counters regression, and stop console errors when MP can't read header

OsamaSayegh:
This PR addresses the issues mentioned in https://github.com/MiniProfiler/rack-mini-profiler/issues/406:

1. Mini Profiler tries to read the `X-MiniProfiler-Ids` header from every `XMLHttpRequest` made on the page. If a request is made to a different origin and the `X-MiniProfiler-Ids` doesn't exist in the response or is not exposed via `Access-Control-Expose-Headers`, an error message will be printed to the console. While this error doesn't affect Mini Profiler's functionality in anyway, it's best to avoid generating noise in the console. This PR amends Mini Profiler behavior so that it only reads the `X-MiniProfiler-Ids` from requests if they're made to the same origin AND the header actually exists.

2. Custom timers aren't showing up in the UI because we regressed when we migrated from jQuery templating to doT.js

3. Another jQuery removal regression that broke the `max_traces_to_show` config.


SamSaffron:
Looks good to me!

PR #429, opened 2020-02-05
Prepend Net::HTTP patch instead of class_eval and aliasing

s-mage:
This patch fixes rack-mini-profiler working along with [http_logger](https://github.com/railsware/http_logger).

Both gems patch `Net::HTTP#request`, but `http_logger` does it using `prepend`, and `rack-mini-profiler` does old-school `class_eval`. 

Problem is that when both gems are on I get `stack overflow` errors, because calling `request_without_mini_profiler` caused calling `request_with_mini_profiler` and so forth.

I've changed `Net::HTTP` patch to use `prepend` as [http_logger does](https://github.com/railsware/http_logger/blob/master/lib/http_logger.rb#L204). It fixed the problem.

Prepend was added in ruby 2.0, and readme says that the gem supports 2.3+, so it's safe to use it.

### DIFF
--- a/lib/patches/net_patches.rb
+++ b/lib/patches/net_patches.rb
@@ -2,7 +2,7 @@
 
 if (defined?(Net) && defined?(Net::HTTP))
   module NetHTTPWithMiniProfiler
-    def request(request, body = nil, &block)
+    def request(request, *args, &block)
       Rack::MiniProfiler.step("Net::HTTP #{request.method} #{request.path}") do
         super
       end
--- a/lib/patches/net_patches.rb
+++ b/lib/patches/net_patches.rb
@@ -1,16 +1,13 @@
 # frozen_string_literal: true
 
 if (defined?(Net) && defined?(Net::HTTP))
-
-  Net::HTTP.class_eval do
-    def request_with_mini_profiler(*args, &block)
-      request = args[0]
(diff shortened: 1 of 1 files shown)


lib/patches/net_patches.rb
```diff
@@ -1,16 +1,13 @@
 # frozen_string_literal: true
 
 if (defined?(Net) && defined?(Net::HTTP))
-
-  Net::HTTP.class_eval do
-    def request_with_mini_profiler(*args, &block)
-      request = args[0]
+  module NetHTTPWithMiniProfiler
+    def request(request, body = nil, &block)
```
SamSaffron: I kind of like the *args thing for safety in case more params are added down the line. Maybe keep that and we can go with the prepend.

s-mage: done. Kept both `request` and `*args` not to do `request = args[0]`

SamSaffron:
Thanks! looking good.

OsamaSayegh:
I'm so sorry @s-mage, but we had to revert this, see https://github.com/MiniProfiler/rack-mini-profiler/commit/2f321c1e3dbfdabb6bed2f9e3af14af2913efffb and https://github.com/MiniProfiler/rack-mini-profiler/issues/437 😢

PR #430, opened 2020-02-07
gemspec: Point out homepage with HTTPS

olleolleolle:
This PR uses `https:` for the homepage URI in the gemspec.


SamSaffron:
Looks good to me!

PR #434, opened 2020-02-19
Fixing getResponseHeader for ancient ExtJS

zaratan:
In old versions of ExtJS, `getResponseHeader` is an object and not a function. Adapting the code to support both.

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -818,7 +818,11 @@ var MiniProfiler = (function() {
           return;
         }
 
-        var stringIds = xhr.getResponseHeader("X-MiniProfiler-Ids");
+        if (typeof xhr.getResponseHeader === 'object') {
+          var stringIds = xhr.getResponseHeader["x-miniprofiler-ids"];
+        } else {
+          var stringIds = xhr.getResponseHeader("X-MiniProfiler-Ids");
+        }
 
         if (stringIds) {
           var ids = stringIds.split(",");


denispasin:
I don't know why Travis is failing though… Should I investigate or is it already in progress?

denispasin:
Hi @Sirlupinwatson1 :)

We can't call
```js
var stringIds = xhr.getResponseHeader("X-MiniProfiler-Ids");
```
because for old `ExtJS` versions `getResponseHeader` is not function and so will fail on that line,

This branch of the if/else: ```js
var stringIds = xhr.getResponseHeader["x-miniprofiler-ids"];
```
has been tested on our side and it works as expected (And yes the case changed between versions)

SamSaffron:
How old is this version of extjs? I am not entirely sure I would like to carry a patch like this. 

The standard here is well established: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader 

Perhaps instead a patch to ensure we get a proper browser implemented XMLHttpRequest object?

denispasin:
The version we seem to have is 1.1.1 (which seems to be from sometime in 2007).

How would you patch this? (I'm asking because I'm unsure of what you are suggesting). Is it: skip it completely if `getResponseHeader` isn't a function?

I completely understand if you don't want to carry a patch for an old ExtJS version. We can keep our fork if needed.

SamSaffron:
I guess if I was stuck maintaining a legacy app and was forced into the unfortunate state of maintaining ancient extjs, maybe I would load a file prior to extjs that salvages XMLHttpRequest and introduce some method of getting one without extjs decorations, inject that in to mini profiler? 

It really is not easy. 

Does that version have a noconflict mode? maybe you can use that so it does not destroy current specifications?

SamSaffron:
Also @denispasin , :hugs:  , that is one pickle you have been placed in :( wishing you much luck convincing the power that be to upgrade stuff. You can site probably 100 CVEs that were filed against old extjs, maybe that can give you a bit of a better fighting chance?

denispasin:
> Does that version have a noconflict mode? maybe you can use that so it does not destroy current specifications?

I don't know and I'll look this way. Thanks a lot for the suggestion.

Regarding the upgrade, we are retiring the current ExtJS UI in favour of a "modern" stack. I've already conducted an upgrade from Rails `1.2.6` forked 10 years ago (with 400 commits on top) to vanilla `2.3.18` this year. I'm not sure I want to engage in another fearful one soon.

nateberkopec:
I think this is stale. Closing.

PR #436, opened 2020-03-12
Skip "/packs" when serving static assets

LynxEyes:
When using `webpacker`, assets are under `/packs` but the `app.config.assets.prefix` still points to `/assets` (because it still exists anyway).

With this we'd stop profiling whatever we'd be serving under `/packs` when serving static assets.

@nateberkopec ☝️

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,22 @@ If you don't want to manually require Mini Profiler:
 gem 'rack-mini-profiler', require: ['enable_rails_patches', 'rack-mini-profiler']
 ```
 
+#### `Net::HTTP` stack level too deep errors
+
+If you start seeing `SystemStackError: stack level too deep` errors from `Net::HTTP` after installing Mini Profiler, this means there is another patch for `Net::HTTP#request` that conflicts with Mini Profiler's patch in your application. To fix this, change `rack-mini-profiler` gem line in your `Gemfile` to the following:
+
+```ruby
+gem 'rack-mini-profiler', require: ['prepend_net_http_patch', 'rack-mini-profiler']
+```
+
+If you currently have `require: false`, remove the `'rack-mini-profiler'` string from the `require` array above so the gem line becomes like this:
+
+```ruby
+gem 'rack-mini-profiler', require: ['prepend_net_http_patch']
+```
+
+This conflict happens when a ruby method is patched twice, once using module prepend, and once using method aliasing. See this [ruby issue](https://bugs.ruby-lang.org/issues/11120) for details. The fix is to apply all patches the same way. Mini Profiler by default will apply its patch using method aliasing, but you can change that to module prepend by adding `require: ['prepend_net_http_patch']` to the gem line as shown above.
+
 #### Rails and manual initialization
 
 In case you need to make sure rack_mini_profiler is initialized after all other gems, or you want to execute some code before rack_mini_profiler required:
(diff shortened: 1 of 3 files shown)


lib/mini_profiler_rails/railtie.rb
```diff
@@ -30,6 +30,7 @@ def self.initialize!(app)
 
     if serves_static_assets?(app)
       c.skip_paths << app.config.assets.prefix
+      c.skip_paths << "/packs"
```
LynxEyes: This is a rather silly solution for the problem..
A better solution here would be to find out if the app is using webpacker or not and add this if yes.

Probably, and hopefully, there is something within webpacker that would give me that "/packs" path without me needing to hardcode here... 

HELP!

nateberkopec: Would `if defined?(Webpacker)` work?

LynxEyes: Would help indeed.

A simple solution with that could be:

```ruby
c.skip_paths << "/packs" if defined?(Webpacker)
```

Problem is: You can tell Webpacker to put those anywhere else, like this:

image

Webpacker allows me to get that path with:

```ruby
Webpacker.config.public_output_path # => #<Pathname:/app/public/packs/like/this>
```

So, I am thinking this needs to be a bit smarter and this is where I start to find my solution ugly:

```ruby
if defined?(Webpacker)
  public_packs_folder = 
    Webpacker.config.public_output_path.to_s.gsub!("#{app.config.root}/public", "")
  c.skip_paths << public_packs_folder
end
```

Is there a cleaner way to do that "substring"?

SamSaffron:
I totally support this, but there is an element of fragility, can we add webpacker to the test suite and have a test to confirm this indeed continues to work?

nateberkopec:
Probably would just need to add/require Webpacker and set those methods correctly.

LynxEyes:
I'm a bit unsure how to add Webpacker to this..

I can't see any other tests being made to the railtie code itself, am I missing something?

OsamaSayegh:
@LynxEyes testing a Railtie/initializer directly is very tricky if not impossible, but what you can do is extract your code into a small method, put it in `railtie_methods.rb` and simply call the method in our Railtie. Then in `railtie_methods_spec.rb` you can call your method as many times as you want and do your tests. You can see https://github.com/MiniProfiler/rack-mini-profiler/commit/b15fe68456ff301616b652fd9a6d097c21392d38 for an example.

Adding Webpacker _should_ be as simple as adding it to the gemspec file as a development dependency, but there may be some hidden complexities 🐉 😄 

If you want, I can pick this up and try to bring it over the finish line.

LynxEyes:
@OsamaSayegh if you have a solid idea of how to finish this, then yes, I fully accept the help! Otherwise I'll follow your tip and try to improve this.

OsamaSayegh:
Lol, I updated your branch to latest and github closed the PR 😆 I'm working on it.

OsamaSayegh:
Actually it looks like I pushed my local master to your fork's master, which essentially undid your commits and github closed the PR because your fork became identical and I don't think I have push permissions anymore because the PR is closed 🤦‍♂️ Silly me, sorry about that!

Anyway, I created a new PR https://github.com/MiniProfiler/rack-mini-profiler/pull/447 with your changes and a test case on top. Will leave to @SamSaffron to do the final review. Thank you for your work here @LynxEyes and sorry for my mess.

LynxEyes:
Lolol. Let that who hasn't messed up a few git merges cast the first stone.
Cool, and thank you for adding the test, I'll take it as reference for future commits I may end up making 🤗

Issue #440, opened 2020-05-18
Regression: client_timings UI, that contained 'show trivial' link, disappeared when upgrading from 1.0.2 to 1.1.6


feliperaul:
Before the upgrade, on v. 1.02, I had this section in my profile that allowed me, amongst other details, to click 'show trivial':

![image](https://user-images.githubusercontent.com/743879/82237595-497ff000-990c-11ea-8a84-eb46006126f6.png)

That square in the picture is a `<div class="profiler-timings profiler-client-timings">`.

However, after the upgrade to 1.1.6, that whole area disappeared:

![image](https://user-images.githubusercontent.com/743879/82237936-d460ea80-990c-11ea-97a8-687e790ce101.png)

Even setting `Rack::MiniProfiler.config.show_trivial = true` doesn't show it back.

SamSaffron:
Confirmed ... this has gone missing @OsamaSayegh  can you have a look?

SamSaffron:
Note ... while looking at this... timing client side have gotten better over the years, there may be some improvements lurking.

feliperaul:
@SamSaffron @OsamaSayegh 
I wanted to point out that the results in the latest version already include the trivial steps in the resulting HTML, but they are hidden by a `display: none` CSS, and there's no link anywhere to be seen to toggle them visible (as there was in the 1.0.2 version).

If I manually disable that CSS rule I get the result I'm looking for (the trivial steps show up). 

![image](https://user-images.githubusercontent.com/743879/82451102-eb761880-9a83-11ea-836c-0d0c36c66e3a.png)

OsamaSayegh:
OK, thanks for reporting this @feliperaul. https://github.com/MiniProfiler/rack-mini-profiler/pull/441 should bring back client timings and the `show trivial` button. I've included a screenshot in the PR. Let me know if there is anything else missing.

OsamaSayegh:
FYI @feliperaul version 2.0.2 is out with this fix 🎉 Thanks for your report!

feliperaul:
Thank you @OsamaSayegh and @SamSaffron !

PR #441, opened 2020-05-21
FIX: Fix client timings regression and bring back 'show trivial'

OsamaSayegh:
Fix for https://github.com/MiniProfiler/rack-mini-profiler/issues/440

Client timings regressed when we removed jQuery a while ago; this PR fixes the regression. Tested on IE11 to ensure MP doesn't regress on IE11.

image


SamSaffron:
cool I will do a new release!!

PR #443, opened 2020-05-22
DEV: Quality of life improvements when working on MP client-side code

OsamaSayegh:
Testing changes to Mini Profiler client-side code is very painful at the moment. When I worked on https://github.com/MiniProfiler/rack-mini-profiler/pull/441 (and other PRs), I went about testing my changes by making my local Discourse instance point to my local clone of mini profiler, and then starting Discourse server to preview my changes. Every time I wanted to preview a change, I had to shut down the Discourse server, run the `update_asset_version` rake task, start the server again and wait for the page to load (which can take a few seconds). Not fun!

So, I decided to scratch my own itch and I've come up with this PR. Taking inspiration from Logster, we now have a new rake task called `client_dev` which will do 2 things:

1. Start a local Sinatra server that serves a simple web page with Mini Profiler. I've also saved a sample `page_struct` from loading a topic in my local Discourse instance, modified it to make it use as many UI elements as I can think of (slow queries, customer timers etc.) and included it in the web page. This should make it possible to test just about everything when making changes to client-side code.

2. Listen to files in the `lib/html` directory and automatically call the `update_asset_version` rake task whenever a file changes. This will make it possible to preview changes by simply reloading the page.

This addresses the biggest pain points I've experienced, and I hope it will help other people!


SamSaffron:
This is great!

PR #444, opened 2020-06-09
DEV: Allow Net::HTTP#request patch to be applied with module prepend or method aliasing

OsamaSayegh:
@SamSaffron what do you think about this? This will allow people to choose how our Net::HTTP is applied like we allow our Rails patches to be optional. People who want our Net::HTTP patch to be applied using module prepend, they can do something like this in their Gemfile:

```ruby
gem 'rack-mini-profiler', require: ['prepend_net_http_patch']
```

Right now people who have other gems with patches that conflict with our patch are forced to either pin their Mini Profiler version and avoid updates, or remove one of the conflicting gems. I think it's nice if we can do something simple to avoid picking sides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,22 @@ If you don't want to manually require Mini Profiler:
 gem 'rack-mini-profiler', require: ['enable_rails_patches', 'rack-mini-profiler']
 ```
 
+#### `Net::HTTP` stack level too deep errors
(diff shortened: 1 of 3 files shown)


SamSaffron:
yes, this seems completely reasonable, be sure you update the readme though

PR #446, opened 2020-06-11
Use redis.exists?

mylesboone:


SamSaffron:
This is not going to work though on earlier redis gems.

maybe cache a `.respond_to?(:exists?)` or something like that to ensure this continues to work on earlier redis.

SamSaffron:
one sec here ... is `redis.exists` going to return non nil for keys that are missing? seems like a massive breaking change to redis. 

If this is not the case, not sure why we would change anything here.

mylesboone:
I'd imagine that `redis.exists` will return a falsey value if the key doesn't exist. Using `redis.exists?` is more intent-revealing and would prevent deprecation warnings for users on `redis-4.2`, the current version.

SamSaffron:
Yeah, we simply can not merge this until 4.2 is almost exclusively used ... adding a respond to here would be message for something that is not going to really change the underlying behavior

PR #450, opened 2020-06-16
Remove Redis 4.2 warnings.

tgxworld:
Apps on Redis 4.2 and up will see a warning due to https://github.com/redis/redis-rb/commit/325752764995b02f17c3e5240ea489f641911d7d

### DIFF
--- a/lib/mini_profiler/storage/redis_store.rb
+++ b/lib/mini_profiler/storage/redis_store.rb
@@ -33,7 +33,7 @@ module Rack
 
       def set_unviewed(user, id)
         key = user_key(user)
-        if redis.exists(prefixed_id(id))
+        if redis.call([:exists, prefixed_id(id)]) == 1
           expire_at = Process.clock_gettime(Process::CLOCK_MONOTONIC).to_i + redis.ttl(prefixed_id(id))
           redis.zadd(key, expire_at, id)
         end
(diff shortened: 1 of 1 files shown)


SamSaffron:
@casperisfine what should our approach be here. 

Our dependency on redis is "weak" in that there is no explicit dependency. `exists?` may no exist in earlier redis gems. 

Should we monkey patch it in and keep this code?

tgxworld:
We can just implement `exists?` since it isn't hard. Essentially just calling `@client.call([:exists, key])`

SamSaffron:
as long as it is backwards compatible with reasonable redis gem releases for whatever definition of reasonable that we can come up with

tgxworld:
We're only using a single key so it should be backwards compatible all the way

casperisfine:
> Should we monkey patch it in and keep this code?

Yeah, based on your constraints I'd do this: https://github.com/MiniProfiler/rack-mini-profiler/commit/44dee3dc008e9addc1b029178980d1bc71be90f2

If the `redis_store.rb` file was autoloaded, you could go with a slightly cleaner alternative:

```ruby
unless Redis.method_defined?(:exists?)
  module RedisExistsRefinement
    refine Redis do
      def exists?(key)
        exists(key)
      end
    end
  end
  using RedisExistsRefinement
end
```

renchap:
Would it be possible to have a release that includes this PR? This is currently preventing our upgrade to the Redis 4.2
Thanks!

SamSaffron:
@renchap sure, just pushed a release.

renchap:
Amazing thanks!

PR #454, opened 2020-07-14
Fix specs.

tgxworld:


tgxworld:
@OsamaSayegh This should fix the specs. Odd that CI doesn't run on PRs here?

OsamaSayegh:
@tgxworld thank you for this! I'll see what's wrong with CI and fix it 👍

Issue #60, opened 2014-03-05
Give an option to precompile includes.css


Crystark:
Hi,

In our production environment, the following file returns a 404: `/mini-profiler-resources/includes.css`

The reason to this is that all our static resources (css, js, images...) are managed by a content delivery server that catches the requests based on the extension of the file and fetches the file directly on disc to optimize the delivery of those.
As `includes.css` isn't actually in the assets directory, it fails to fetch it.

Would it be possible to give an option to allow this file to be precompiled alongside the other CSS files when using `rake assets:precompile` ?

Thanks

pencilcheck:
+1

5minpause:
+1
Is there any development for this issue?
Would it be an option to include the .css files in the assets pipeline if present?

stiller-leser:
Has there been any work on this issue? We are facing the same issue when trying to integrate the profiler into an Rails-app with asset-pipeline. It won't get the include.js - had a look at gemifying it, but there are a lot of changes needed. Hard to do, if you don't know the code status.

SamSaffron:
totally open to allowing the assets to be included via sprockets, it should be fairly straight forward to just move stuff around a bit to make this workable no?

factcondenser:
I believe this issue has been addressed per the merged PR mentioned above ☝️. @SamSaffron - should be safe to close this issue unless anyone objects?

SamSaffron:
thanks @factcondenser closed.

PR #456, opened 2020-07-14
Allow assets to be precompiled with Sprockets

OsamaSayegh:
Context https://github.com/MiniProfiler/rack-mini-profiler/issues/60.

This makes it possible to consume Mini Profiler's assets like any other assets in a Rails app. There is a new config option `assets_url` which is `nil` by default but can be changed to a lambda that takes 3 arguments: asset_name (i.e. `rack-mini-profiler.{js|css}`), version (which is the digest/version of MP assets), and `env` object, and the lambda should return a string that will be used as the `src` for the asset. To preserve backward compatibility, the lambda is optional and Mini Profiler will fallback to the current behavior (serve assets from `/mini-profiler-resources/*`) if no lambda is given or it returns a falsey value.

For most Rails app, the following example will work fine:

```
Rack::MiniProfiler.config.assets_url = -> (name, version, env) do
  ActionController::Base.helpers.asset_url(name)
end
```

I tested this change in a brand new Rails app and Discourse and in both apps `rake assets:precompile` generated MP assets in the `public/assets` directory, and the assets loaded just fine from the `/assets` route when I loaded a page in my browser.

Thoughts on this @SamSaffron? Is this OK? Anything you would like me to change or test?


SamSaffron:
I think I am fine as long as this is 100% opt in, the PR appears to adding stuff to sprockets by default, we should only add it if people opted to sprockets management.

lib/mini_profiler_rails/railtie.rb
```diff
@@ -6,6 +6,13 @@
 module Rack::MiniProfilerRails
   extend Rack::MiniProfilerRailsMethods
 
+  class Engine < ::Rails::Engine
```
SamSaffron, line 9: this whole thing should be opt in I think.

OsamaSayegh:
Ok, I've changed this so that the engine is only created if the `assets_url` config is set. If this looks good I'll add documentation to the README file and merge it in.

OsamaSayegh:
I'm going to merge this tomorrow unless there are objections @SamSaffron 😸

SamSaffron:
looks good to me!

PR #461, opened 2020-09-01
FEATURE: Allow skip_paths config to contain regular expressions

OsamaSayegh:
@SamSaffron when I was trying to enable snapshots on Discourse, I noticed we don't use Mini Profiler's `skip_paths`
 config but instead use `pre_authorize_cb` to skip paths:

https://github.com/discourse/discourse/blob/90eeb8f7d95bf8e06adb34acbbb65e2c743e551f/config/initializers/006-mini_profiler.rb#L53-L58

We also use `pre_authorize_cb` to disable profiling on phones/tablets.

Skipping paths in `pre_authorize_cb` is problematic for snapshotting because requests that fail the `pre_authorize_cb` check are eligible for snapshotting, and we do want requests from phones/tablets to be eligible for snapshotting, but we never want (I think) requests to skipped paths to be eligible for snapshotting (e.g. I don't think we want snapshots for assets requests).

The only thing stopping us from using `skip_paths` is that we want regular expressions but `skip_paths` currently accepts strings only. This PR allows both strings and regular expressions in `skip_paths`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ Option|Default|Description
 -------|---|--------
 pre_authorize_cb|Rails: dev only<br>Rack: always on|A lambda callback that returns true to make mini_profiler visible on a given request.
 position|`'top-left'`|Display mini_profiler on `'top-right'`, `'top-left'`, `'bottom-right'` or `'bottom-left'`.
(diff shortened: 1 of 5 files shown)


SamSaffron:
this seems fine to me... thanks!

PR #463, opened 2020-09-23
FEATURE: Snapshots transporter

OsamaSayegh:
This feature can be used to ship snapshots from multiple sites to a central place where all snapshots can be viewed from a single place.


lib/mini_profiler/profiler.rb
```diff
@@ -809,10 +814,14 @@ def take_snapshot(env, start)
         )
         custom_fields = MiniProfiler.get_snapshot_custom_fields
         page_struct[:custom_fields] = custom_fields if custom_fields
-        @storage.push_snapshot(
-          page_struct,
-          @config
-        )
+        if Rack::MiniProfiler.snapshots_transporter?
```
OsamaSayegh, line 817: @SamSaffron should we skip storage if transportation is configured? This line will skip storage if transportation is configured, but it shouldn't be difficult to change it so we store snapshots and ship them. A downside of skipping storage is that some snapshots might get lost if ruby process is killed/restarted before buffered snapshots are shipped.

SamSaffron: Yeah I think we can skip for now if transport is configured, seems reasonable.

lib/mini_profiler/snapshots_transporter.rb
```diff
@@ -0,0 +1,64 @@
...
+  def ship(snapshot)
+    @buffer_mutex.synchronize do
+      @buffer << snapshot
+      @buffer.shift if @buffer.size > @max_buffer_size
+    end
+    @thread_mutex.synchronize { start_thread }
+  end
+
+  def flush_buffer
+    if @buffer.size > 0
```
SamSaffron, line 34: I think you need to protect access to @buffer with a mutex, this will technically work on MRI but can fail on JRuby

lib/mini_profiler/snapshots_transporter.rb
```diff
@@ -0,0 +1,64 @@
...
+      request = Net::HTTP::Post.new(
+        @uri,
+        'Content-Type' => 'application/json',
+        'Mini-Profiler-Transport-Auth' => @auth_key
+      )
+      request.body = { snapshots: buffer_content }.to_json
+      http = Net::HTTP.new(@uri.hostname, @uri.port)
+      http.use_ssl = @uri.scheme == 'https'
+      res = http.request(request)
+      if res.code.to_i == 200
```
SamSaffron, line 45: we should probably at least expose some top level metrics here under mini profiler namespace for 

- Number of things we shipped successfully 
- Number of 200 responses
- Number of unknown errors  

attr_readers on this class is probably good enough then we can dig this out using prometheus which is hand for debugging issues.

SamSaffron:
Overall a great change and well designed

OsamaSayegh:
Ok I've made the changes you asked. Let me know if there is anything else 🙂

SamSaffron (approved):

SamSaffron:
Looking good to me, merge away!

PR #464, opened 2020-09-24
Fix sameSite cookie error in console

### LINKED ISSUE #442: sameSite warning for __profilin cookie in browser console
With Rails 4.2 and rack-mini-profiler 2.0.1 I get in my browser console:

```
Cookie “__profilin” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies
```

I was able to remove this warning for my main cookie for the app via:

```
Rails.application.config.session_store :cookie_store, key: '_quepid_session', same_site: :lax 
```

Thoughts on being able to control the sameSite setting?  Or providing a default?

westonganger:
Fixes #442

### DIFF
--- a/lib/mini_profiler/client_settings.rb
+++ b/lib/mini_profiler/client_settings.rb
@@ -76,6 +76,7 @@ module Rack
           settings_string = settings.map { |k, v| "#{k}=#{v}" }.join(",")
           cookie = { value: settings_string, path: '/', httponly: true }
           cookie[:secure] = true if @request.ssl?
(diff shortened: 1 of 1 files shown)


SamSaffron:
@OsamaSayegh  this looks safe to me, should we merge?

OsamaSayegh:
Yes, sure. Thanks @westonganger !

PR #466, opened 2020-10-02
Don't store SQL queries for snapshots

OsamaSayegh:
This is done for security reasons since SQL queries may contain secrets
that shouldn't be shared around.

cc @SamSaffron


lib/mini_profiler/profiling_methods.rb
```diff
@@ -7,7 +7,14 @@ module ProfilingMethods
       def record_sql(query, elapsed_ms, params = nil)
         return unless current && current.current_timer
         c = current
-        c.current_timer.add_sql(query, elapsed_ms, c.page_struct, params, c.skip_backtrace, c.full_backtrace)
+        c.current_timer.add_sql(
+          is_snapshot? ? nil : query,
```
SamSaffron, line 11: I think we make this optional, by default we don't ship queries or env, but if you opt for it you can get queries and even env.

OsamaSayegh: I've added an config to keep SQL queries in snapshots, but I'm not sure what you're referring to by "env". Can you clarify what you mean by env please?

SamSaffron: oh ... yeah I got totally confused with ?pp=env (which is not glued to each request) ... yeah config to keep sql is all we need, and an update to the readme.

SamSaffron (approved):

PR #468, opened 2020-11-04
FEATURE: Exponential backoff for snapshots transporter

OsamaSayegh:
The snapshots transporter should wait more between requests if the configured destination is not working as expected.


SamSaffron:
Nice well tested change!

SamSaffron (approved):

PR #469, opened 2020-11-05
FEATURE: Add optional gzip compression to snapshots transporter requests

OsamaSayegh:
This adds optional gzip-compression to requests made by the snapshots transporter. It's off by default and can be changed via the new `snapshots_transport_gzip_requests` config.


SamSaffron:
can we just do this all transparently with content-encoding: gzip on the request?  

https://serverfault.com/questions/56700/is-it-possible-to-enable-http-compression-for-requests

OsamaSayegh:
Sorry I'm not sure I follow. Are you asking if it's possible to avoid doing the compression ourselves and let something else handle it by simply indicating we want compression via the `content-encoding: gzip` header?

lib/mini_profiler/snapshots_transporter.rb
```diff
@@ -50,12 +51,24 @@ def flush_buffer
...
-      request.body = { snapshots: buffer_content }.to_json
+      }
+      json = { snapshots: buffer_content }.to_json
+      body = if @gzip_requests
+        require 'zlib'
+        io = StringIO.new
+        gzip_writer = Zlib::GzipWriter.new(io)
+        gzip_writer.write(json)
+        gzip_writer.close
+        headers['Content-Encoding'] = 'gzip'
```
SamSaffron, line 65: oh I missed this line...  This change looks fine to me, I do wonder though if we need this toggle `snapshots_transport_gzip_requests` ... I guess it teaches people that you can compress.

Approving!

SamSaffron (approved):

PR #473, opened 2020-12-17
prettier rb

SamSaffron:
- somehow my sass generator is a little bit off each time
- Work-in-progress, enforce formatting via prettier rb


nateberkopec:
Closing as stale, but opening an issue re: formatter.

Issue #467, opened 2020-10-08
When Turbolinks is enabled rack-mini-profiler makes and ever increasing number of duplicated XHR requests for profiler results


jimeh:
I believe there's an issue around JavaScript event handlers which register themselves on page load to fetch profiler results from `/mini-profiler-resources/results`. Each time a Turbolinks link is clicked, it seems to create another results fetcher handler, leading to an ever increasing number of duplicated handlers making identical requests to `/mini-profiler-resources/results`.

I have a basic proof of concept app:

- URL: https://cryptic-dawn-20880.herokuapp.com/hello/page_a
- Source: https://github.com/jimeh/rack-mini-profiler-bug-demo

It's a blank Rails 6 app, two pages with links back and forth between them, and rack-mini-profiler added.

And here's a gif of if too for good measure:

![profiler-bug](https://user-images.githubusercontent.com/39563/95492162-94210800-0992-11eb-9cb7-a7d07c3d72ef.gif)

zokioki:
Can confirm this is happening on my Turbolinks-enabled app as well (Rails 6.1.0, Turbolinks 5.2.0).

SamSaffron:
@OsamaSayegh can you investigate?

OsamaSayegh:
I've opened https://github.com/MiniProfiler/rack-mini-profiler/pull/474 to fix this. Thank you so much @jimeh for the proof of concept, it was really helpful for debugging this issue 👍

SamSaffron:
@zokioki / @jimeh  can you confirm it corrects the issue? If so I will push a new version of the gem out there.

zokioki:
@SamSaffron Confirming that this does fix the issue in my case - thanks y'all! 👍

jimeh:
Sorry for my delayed response.

I can confirm it has fixed the issue on my end as well. I've updated the demo Rails app as proof too: https://cryptic-dawn-20880.herokuapp.com/hello/page_a

jimeh:
@OsamaSayegh thanks for the fix :)

PR #474, opened 2020-12-22
FIX: Fetch Mini Profiler results only once in Turbolinks apps

OsamaSayegh:
Fix for https://github.com/MiniProfiler/rack-mini-profiler/issues/467.


SamSaffron:
Cool lets give this a shot!

PR #475, opened 2020-12-25
FEATURE: Use speedscope to display flamegraphs and switch from the flamegraph gem to stackprof

OsamaSayegh:
Mini Profiler now ships with an embedded copy of speedscope which
provides a very nice experience for viewing flamegraphs.

Flamegraph has been replaced with the stackprof gem. It's an optional
dependency, so if you use the flamegraphs feature, you'll need to add
the `stackprof` gem to your `Gemfile` and require it. If you don't use
the feature, you don't need to add anything and Mini Profiler will work
just fine.

image


SamSaffron:
Looks great to me :confetti_ball:  

Let's add a rake task to update us to latest speedscope `rake update:speedscope` that way we can get upstream fixes easily.

jlfwong:
Hi! I'm the author of speedscope and just wanted to say that this is awesome! I made an opinionated fork of rack-mini-profiler in the past to do this (https://github.com/jlfwong/rack-mini-profiler), but so happy this is in `rack-mini-profiler` core now!

SamSaffron:
Hi Jamie, huge fan! Thanks so much for building speedscope

PR #478, opened 2020-12-31
DEV: Add rake task for upgrading speedscope

OsamaSayegh:
This also adds a local copy of the Source Code Pro font (which is the only third-party asset that speedscope loads) to make sure speedscope works when you're behind an intranet (the rake task takes care of making sure index.html links to the local copy when doing upgrades).


SamSaffron:
Nice one! thanks!

Issue #480, opened 2021-01-12
ProfilingMethods in Ruby 3.0.0


kaiquekandykoga:
I opened an issue recently for Ruby on Rails https://github.com/rails/rails/issues/41069, however I was advised to open an issue here instead.

ActionView uses this part of code:

https://github.com/MiniProfiler/rack-mini-profiler/blob/c7742152bae5f815e5e5da3a1d095889cac3e573/lib/mini_profiler/profiling_methods.rb#L84

When args contains `[#<ActionView::Base:0x0000000000a050>, {}, {:add_to_stack=>true}]`

It returns args to ActionView here:

https://github.com/MiniProfiler/rack-mini-profiler/blob/c7742152bae5f815e5e5da3a1d095889cac3e573/lib/mini_profiler/profiling_methods.rb#L111

ActionView is receiving these values in this signature:

`def render(view, locals, buffer = ActionView::OutputBuffer.new, add_to_stack: true, &block)`

https://github.com/rails/rails/blob/19759dfaf38406907dd360f29527152d15c5e6e5/actionview/lib/action_view/template.rb#L150-L158

Using Ruby 3.0.0 buffer captures `{:add_to_stack=>true}` from args causing an exception later.

The idea given to me was to use **kargs. https://github.com/rails/rails/issues/41069#issuecomment-758069051

Thank you a lot.

SamSaffron:
@OsamaSayegh is working on a fix, we should have a release soon.

OsamaSayegh:
Hi @kaiquekandykoga,

version `2.3.1` is released and it includes a fix to this problem. Thanks for reporting this to us!

kaiquekandykoga:
Thank you very much.

PR #481, opened 2021-01-25
Make MiniProfiler.profile_method compatible with Ruby 3

OsamaSayegh:
Fix for https://github.com/MiniProfiler/rack-mini-profiler/issues/480.

We can't pass `**kwargs` unconditionally because it would break backward compatibility with ruby < 2.7:

```
~ » cat ~/test.rb
class A
  def t(a, b = 2)
    { a: a, b: b }
  end
end

A.send :alias_method, 'old_t', 't'
A.send :define_method, 't' do |*args, **kwargs, &blk|
  self.send 'old_t', *args, **kwargs, &blk
end

puts RUBY_VERSION
puts A.new.t(10)

~ » RBENV_VERSION=2.7.0 rbenv exec ruby ~/test.rb
2.7.0
{:a=>10, :b=>2}  # <=== 👍 

~ » rbenv exec ruby ~/test.rb
2.6.6
{:a=>10, :b=>{}} # <=== 👎 
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,11 @@ jobs:
         options: --health-cmd "timeout 5 bash -c 'cat < /dev/null > /dev/udp/127.0.0.1/11211'" --health-interval 10s --health-timeout 5s --health-retries 5
     strategy:
       matrix:
-        ruby: ["2.7", "2.6", "2.5"]
+        ruby: ["3.0", "2.7", "2.6", "2.5"]
         redis: ["5.x"]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


SamSaffron:
I think cause this is so perf sensitive and we are simply forwarding calls we should use ruby2_keywords 

https://bugs.ruby-lang.org/issues/16897

then longer term once we only target Ruby 3 we can start using ... delegation.

OsamaSayegh:
Ok I've updated this to use `ruby2_keywords`.

SamSaffron:
PERFECT, nice simple patch :heart: 

Will push out a new version now.

PR #484, opened 2021-02-26
Remove obsolete Ruby version requirement notes

semaperepelitsa:
This gem already requires Ruby version >= 2.4


SamSaffron:
Looks good to me, thanks!

PR #488, opened 2021-03-31
Add memory_profiler error when not installed

gmcgibbon:
Adds error message when `memory_profiler` isn't installed. The wording and logic is borrowed from the error you get from `pp=flamegraph` when `stackprof` isn't installed. I also added some integration tests for each feature.


SamSaffron:
looks good to me! 

btw 2.5 is now EOL we are going to remove it from the test matrix soon.

PR #491, opened 2021-04-10
Offer "allow_authorized" as an alternative to "whitelist"

rthbound:
- Some teams and projects will wish to use rack-mini-profiler without
  needing to use terminology they feel is exclusive/biased/harmful
  - This change adds support for 'allow_authorized' as a suitable alternative
  for replacing 'whitelist'
  - In truth, the concept of 'explicit authorization' is more true to how this
  library works: there is no list of allowed endpoints, but rather some
  endpoints are explicitly allowed and others are not.
  - No immediate change required for clients -- they can continue to specify whatever
  authorization mode they used before this change. Using the old authorization mode identifier will emit a deprecation warning and set the authorization mode to ":allow_authorized" instead.
  - Include tests showing backward compatibility and support for the new
  authorization mode identifier.
  - Update integration spec to: utilize :allow_authorized authorization mode; bring ancillary language into alignment

### DIFF
--- a/lib/mini_profiler/config.rb
+++ b/lib/mini_profiler/config.rb
@@ -86,6 +86,20 @@ module Rack
 
       attr_reader :assets_url
 
+      def authorization_mode=(mode)
+        if mode == :whitelist
+          warn "[DEPRECATION] `:whitelist` authorization mode is deprecated. Please use `:allow_authorized` instead."
+
+          mode = :allow_authorized
+        end
+
+        warn <<~DEP unless mode == :allow_authorized || mode == :allow_all
+          [DEPRECATION] unknown authorization mode #{mode}. Expected `:allow_all` or `:allow_authorized`.
+        DEP
+
+        @authorization_mode = mode
+      end
+
       def assets_url=(lmbda)
         if defined?(Rack::MiniProfilerRails)
           Rack::MiniProfilerRails.create_engine
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -213,7 +213,7 @@ module Rack
     def call(env)
       start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
       client_settings = ClientSettings.new(env, @storage, start)
-      MiniProfiler.deauthorize_request if @config.authorization_mode == :whitelist
+      MiniProfiler.deauthorize_request if @config.authorization_mode == :allow_authorized
 
       status = headers = body = nil
       query_string = env['QUERY_STRING']
@@ -239,7 +239,7 @@ module Rack
       skip_it = (@config.pre_authorize_cb && !@config.pre_authorize_cb.call(env))
 
       if skip_it || (
-        @config.authorization_mode == :whitelist &&
+        @config.authorization_mode == :allow_authorized &&
         !client_settings.has_valid_cookie?
       )
         if take_snapshot?(path)
@@ -388,7 +388,7 @@ module Rack
 
       skip_it = current.discard
 
-      if (config.authorization_mode == :whitelist && !MiniProfiler.request_authorized?)
+      if (config.authorization_mode == :allow_authorized && !MiniProfiler.request_authorized?)
         skip_it = true
       end
 
--- a/README.md
+++ b/README.md
@@ -235,21 +235,21 @@ rack-mini-profiler is designed with production profiling in mind. To enable that
 
 Note:
 
-Out-of-the-box we will initialize the `authorization_mode` to `:whitelist` in production. However, in some cases we may not be able to do it:
+Out-of-the-box we will initialize the `authorization_mode` to `:allow_authorized` in production. However, in some cases we may not be able to do it:
 
-- If you are running in development or test we will not enable whitelist mode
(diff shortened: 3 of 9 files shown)


lib/mini_profiler/config.rb
```diff
@@ -86,6 +86,10 @@ def self.default
 
       attr_reader :assets_url
 
+      def explicit_authorization_mode?
+        [:explicit, :whitelist].include? authorization_mode
```
SamSaffron, line 90: I see no particular reason for keeping this long term ... when you set authorization_mode to :whitelist we should simply print a deprecation and set it to :explicit. then we can remove the deprecation in a few releases. Then we get to avoid recreating this array all the time and the extra method call.

rthbound: Good idea -- I've added 4b1f3aae9c1c3d3f59b58c1d3bd9451ef043cd7b which takes that approach using `Warning` to emit the deprecation message. Not sure if the latest of this MR will be satisfactory or if we'd prefer something like `!!logger ? logger.warn(msg) : warn(msg)`.

Thanks for the suggestion. Happy to make further adjustments as needed!

SamSaffron:
something about the words here is not working for me: 

We have `#authorization_mode=` , `explicit` and `allow_all`

explicit does not feel like the correct word.

Maybe instead `allow_all` and `allow_authorized`  ? Maybe `allow_all` and `deny_unauthorized` ? 


Also since we have a setter here, we should exception out if people try to set an unexpected authorization mode.

rthbound:
@SamSaffron I did consider throwing an exception if people set an unexpected mode, however this could be a breaking change:

- It's possible users have configured the library with some typo
- They may have passed in strings instead of symbols

In either of these cases, `MiniProfiler.config.authorization_mode == :whitelist` would have returned false, sure, but there would not have been an exception.

What do you think we should do? We could emit a deprecation warning and claim that in the future we will raise if the given argument is not one of `:allow_all` or `:allow_authorized`.

My thinking was that this entire setter method would disappear once `:whitelist` is fully removed.

SamSaffron:
Yeah let's do a deprecation for the next 6 months and then we can remove the deprecation and just throw an exception for unexpected modes.

lib/mini_profiler/config.rb
```diff
@@ -86,6 +86,20 @@ def self.default
...
+      def explicit_authorization_mode?
+        authorization_mode == :allow_authorized
+      end
+
+      def authorization_mode=(mode)
+        if mode == :whitelist
+          warn "[DEPRECATION] `:whitelist` authorization mode is deprecated. Please use `:allow_authorized` instead."
+
+          mode = :allow_authorized
+        end
```
SamSaffron, line 98: let's add one more deprecation here for not :allow_all or :allow_authorized.

lib/mini_profiler/profiler.rb
```diff
@@ -213,7 +213,7 @@ def tool_disabled_message(client_settings)
     def call(env)
       start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
       client_settings = ClientSettings.new(env, @storage, start)
-      MiniProfiler.deauthorize_request if @config.authorization_mode == :whitelist
+      MiniProfiler.deauthorize_request if @config.explicit_authorization_mode?
```
SamSaffron, line 216: I know it is ugly, but we save on one method call in a critical path... I am a bit nit picky here cause this happens on every request in production 

`if @config.authorization_mode == :allow_authorized`

SamSaffron:
A few tiny changes and I think we are good to merge! thanks @rthbound  :confetti_ball:

lib/mini_profiler/profiler.rb
rthbound: Now that you mention it, the predicate method was added primarily to wrap up the original approach for maintaining backward compatibility with clients that are currently specifying `:whitelist`.

I think the `#explicit_authorization_mode?` method is a bit unnecessary now that we've implemented your recommendation to coerce the authorization mode to `:allow_explicit` when given the old identifier.

I'm guessing you'll agree -- I think I'll just refactor it away entirely.

rthbound:
@SamSaffron thank you for your collaboration on this!

lib/mini_profiler/config.rb
```diff
@@ -86,6 +86,24 @@ def self.default
 
       attr_reader :assets_url
 
+      def explicit_authorization_mode?
+        authorization_mode == :allow_authorized
+      end
```
SamSaffron, line 91: lets remove this method, then we are good to merge!

rthbound: Can't believe I missed that -- thank you!

SamSaffron (approved):

SamSaffron:
looking great, merging this in, any idea why CI is playing up?

rthbound:
I haven't seen that before. Looks flaky -- this report of that particular error claims it was happening around ~5% of the time, once upon a time: https://github.com/rubygems/rubygems/issues/4187

SamSaffron:
No worried Ryan, yeah looks like it is passing once merged. 

Thanks again for the contribution. We will move from master -> main as well some time in the next few weeks.

PR #492, opened 2021-04-13
Fix profiler pop-up extending beyond the window, impossible to scroll

semaperepelitsa:
When viewing a request that is not the first, the bottom of the pop-up comes outside of the window. Even scrolling all the way doesn't let you see the bottom of it.

Before/After (blocked out unnecessary info)
Screenshot 2021-04-14 at 00 57 07
Screenshot 2021-04-14 at 00 58 04


SamSaffron:
oh I see, nice fix :+1:

PR #494, opened 2021-04-27
FEATURE: Introduce `pp=async-flamegraph` for asynchronous flamegraphs

davidtaylorhq:
Using `?pp=async-flamegraph` causes the flamegraph data to be placed in long-term storage, and made available via a link in the mini_profiler UI. Flamegraph data will also be recorded and stored for all AJAX requests with `?pp=async-flamegraph` in the `Referer` header. This is useful in a few situations:

- You want to view flamegraphs for AJAX requests made by a Javascript application. By supplying `pp=async-flamegraph`, flamegraph links for every AJAX request will be made available in the mini-profiler UI.

- You want to see the HTML result of a request, and view the flamegraph later. The existing `?pp=flamegraph` option hides the true output.

- You are performing the request via a tool like `curl`, and would like to view the flamegraph later in the browser (you can extract the X-MiniProfiler-Ids header from the response, then view flamegraph in the browser)

---

When the `pp=async-flamegraph` parameter is supplied, a new "flamegraph" link is added to the UI:

Screenshot 2021-04-27 at 23 27 00

Clicking the link will take you to a URL like `/mini-profiler-resources/flamegraph?id=t0x70kt7hy3cmitx7adx`, which displays the flamegraph UI.


SamSaffron:
I see what you did here! looks great. Some feedback

1. How sticky is Referer? For something like Discourse do we want to clear it or maintain it between interior link? 

2. We need some documentation in the README

3. We need some tests

4. Do we want a "ninja" options of "Please gather me 3 flamegraphs for routes matching `/my/amazing/page` ? This is extremely powerful for debugging and pretty usable.  (similar to how we do sampling) 

@OsamaSayegh can you have a look as well?

davidtaylorhq:
> How sticky is Referer? For something like Discourse do we want to clear it or maintain it between interior link?

Yeah I think we want to make it sticky for internal links in Discourse. I made a draft for that in https://github.com/discourse/discourse/pull/12863

Agreed on 2 & 3 - will look at those things tomorrow.

> Do we want a "ninja" option of "Please gather me 3 flamegraphs for routes matching /my/amazing/page ? This is extremely powerful for debugging and pretty usable. (similar to how we do sampling)

Sounds cool! How do you imagine the UX working here? I guess flamegraphs could be added to the sampling feature, although it is quite a lot of data so we probably wouldn't want to be shipping it to other sites like we do with the other sampled data.

SamSaffron:
> Sounds cool! How do you imagine the UX working?

I guess I would start backwards :) My use case is: 

1. I am on the site `awesome.com` I have MiniProfiler access on the site
2. I would like to get 10 flamegraphs for the route `/homepage` from random users.

Even though I can get a flamegraph today for `/homepage` it will include some extra "admin" and "logged on user" data. This may be less interesting for a particular profiling session.

I am thinking I would use a special route to "enable collection" for a path. Then I would head to another page to see all the current collected results. 

A "list" UI is generally useful anyway for the REFERER solution cause it allows you to easily dig through history.

I would say that data forwarding (like we do for sampling) may be a v3 kind of feature. (list UI and strategic route based collection can also be a v2 thing, it should not hold up this PR)

lib/html/includes.tmpl
```diff
@@ -142,6 +142,9 @@
 <script id="linksTemplate" type="text/x-dot-tmpl">
   <a href="{{= MiniProfiler.shareUrl(it.page.id) }}" class="profiler-share-profiler-results" target="_blank">share</a>
   <a href="{{= MiniProfiler.moreUrl(it.timing.name) }}" class="profiler-more-actions">more</a>
+  {{? it.page.flamegraph}}
```
OsamaSayegh, line 145: This is kind of a nitpick, but I think `it.page.flamegraph` here is the actual flamegraph? Since we don't need the whole thing and just need to know if it exists, maybe we could do some magic when serializing here:

https://github.com/MiniProfiler/rack-mini-profiler/blob/aef9b2308a96be9be2236005d854a0e1d47cad00/lib/mini_profiler/timer_struct/page.rb#L114-L120

and replace `flamegraph` with a boolean `flamegraph_exists` on the fly? That way we don't have to serialize and send flamegraph data (which tends to be very large) to clients.

OsamaSayegh (approved):
Looks excellent to me 👌 Just one minor comment/question.

davidtaylorhq:
Thanks @SamSaffron and @OsamaSayegh. Updated based on your comments, now with tests, and updates to the README/CHANGELOG.

spec/integration/mini_profiler_spec.rb
```diff
@@ -141,6 +141,39 @@ def app
...
+      id = last_response.headers['X-MiniProfiler-Ids'].split(",")[0]
+      get "/mini-profiler-resources/flamegraph?id=#{id}"
+      expect(last_response.status).to eq(404)
+    end
+
+    Process.wait(pid)
+    expect($?.exitstatus).to eq(0)
+  end
+
+
```
OsamaSayegh, line 176:
```suggestion
```

(rubocop complaining about this blank line)

OsamaSayegh:
Ok, I'm merging this in, thanks @davidtaylorhq ❤️ 

We will need @SamSaffron to push a new version on rubygems because I don't think I have push access there.

Issue #439, opened 2020-04-22
Injected content doesn't play well with CSP nonces


viraptor:
When setting up CSP in rails, the mini-profiler is a resource which is hard/impossible to manage properly with script/style nonces. When using domain whitelists, you can add `'self'` to work around the problem, but once you use nonces, the domain whitelist is disabled.

It would be great if there was a setting enabling injecting `nonce="#{content_security_policy_nonce}"` when using mini-profiler with rails.

richhollis:
I've sumitted a PR for this: https://github.com/MiniProfiler/rack-mini-profiler/pull/495

PR #495, opened 2021-05-09
Add support for adding CSP nonce to profile handler JS script tag

richhollis:


richhollis:
Hi @SamSaffron 

I've just pushed some revisions. Please take a look and let me know if this aligns with what you're thinking. If you need anything else then just let me know.

Cheers

lib/mini_profiler/profiler.rb
```diff
@@ -127,6 +127,11 @@ def user(env)
       @config.user_provider.call(env)
     end
 
+    def content_security_policy_nonce(env)
+      config_csp_nonce = @config.content_security_policy_nonce
+      config_csp_nonce.is_a?(String) ? config_csp_nonce : config_csp_nonce.call(env)
```
SamSaffron, line 132: I would prefer to "ultra" optimize this pipeline, can we wipe out the proc here? no need to have this extra indirection here imo: 

A simple thing like this should do the trick, no? 

`@config.content_security_policy_nonce || env["action_dispatch.content_security_policy_nonce"] || env["secure_headers_content_security_policy_nonce"]`

SamSaffron:
looks great to me! thanks :confetti_ball:

richhollis:
Great! :+1: This should close #439.

PR #496, opened 2021-05-14
Inline styles can violate CSP style-src directives

richhollis:
Hello again

When originally testing the last PR for CSP, I overlooked the fact that the project in question used unsafe-inline for styles, so I didn't pick up the issue with inline styles that rack-mini-profiler was using. When switching to a newer project, I could see console errors for CSP violations for inline styles when the DOM was being manipulated in renderTemplate, preventing the script from inserting into the DOM.

With a well defined CSP, you would not normally allow inline styles, so you would need to remove any inline styles to benefit from this protection.

This PR addresses this problem by eliminating the two inline styles and resolves the CSP violation messages.

Let me know if you have any questions.

Thanks
Rich


SamSaffron:
Very nice, super happy to get rid of the inline styles.

richhollis:
> Very nice, super happy to get rid of the inline styles.

That feeling you get, when you purge a few inline styles... :smiley:

PR #497, opened 2021-05-29
Add charset=utf-8 to content-type text/plain

akiomik:
`memory_profiler` reports may contain non-ASCII characters (e.g. Allocated String Report section), so if charset is not specified, the results in `?pp=profile-memory` may be garbled.

Fixed this problem by specifying charset for `Content-Type` of `text_result`.


SamSaffron:
Change looks good, thanks for the fix!

Issue #482, opened 2021-02-17
Missing support for HotWire Turbo Drive, successor of turbolinks


mildred:
When my turbo drive page reloads, the profiler information is removed. This did not happen when I was using turbolinks.

Most things should work the same, just replace `turbolinks` by `turbo`.

mildred:
I tried replacing turbolinks with turbo in includes.js but it is not working. I don't understand how it is supposed to work. As far as I understand the implementation, this is what happens when a turbolink visit happens:

- a visit happens, turbo loads asynchronously the new page and extract the body
- the body is replaced and because the includes.js script is not marked with `data-turbo-permanent="true"`, the script is removed from the page and a new script is inserted
- MiniProfiler include.js script is loaded anew and initialized again
- because upon reinitialization, MiniProfiler does not patch XMLHttpRequest nor window.fetch again, it performs only partial initialization
- XMLHttpRequest and window.fetch are patched using the previous instance of MiniProfiler, and when a query is made, it functions as usual. Only, MiniProfiler is adding profiling results to a container that is disconnected from the page DOM

Over the time, MiniProfiler is initialized many times. I believe that it should be marked with `data-turbo-permanent="true"` to avoid multiple instances of the same script in memory, and conflicting callbacks referring to disconnected instances but still living in some ways. The other thing to do is to reconnect the container to the page body if MiniProfiler detects it is no longer connected (because Turbo removed it from the page)

mildred:
My experiments now involved initialization in rails:

```ruby
  module RackTurboExtension
    def get_profile_script(*args, **kargs)
      super(*args, **kargs).gsub('></script>', ' data-turbo-permanent="true"></script>')
    end
  end

  module Rack
    class MiniProfiler
      prepend RackTurboExtension
    end
  end
```

and in javascript:

```js
document.addEventListener("turbo:before-visit", e => {
  window.MiniProfilerContainer = document.querySelector('body > .profiler-results')
  //window.MiniProfiler.pageTransition()
})

document.addEventListener("turbo:load", e => {
  if(window.MiniProfilerContainer) {
    document.body.appendChild(window.MiniProfilerContainer)
  }
})
```

kbrock:
Bonjour Mildred,

Have you tried putting the code into the actual source code as a PR?

both look pretty straight forwards (though to be honest I'm more of a back end developer than a front end one)

ceritium:
[...]

PR #498, opened 2021-06-11
Add support for Hotwire Turbo Drive

ceritium:
It is related to this issue https://github.com/MiniProfiler/rack-mini-profiler/issues/482

I moved to the gem the @mildred's solution.

I tested it on a small demo app, as I show in the following video: 

https://user-images.githubusercontent.com/16633/121650575-fa1ddc00-ca99-11eb-95eb-fc1b4fb0698d.mov

Any recommendation for better QA?


Dainii:
Tested with rails (6.1.3.1) and turbo-rails (0.5.9), it seems to work quite nicely.

lib/html/includes.js
```diff
@@ -640,6 +640,17 @@ var _MiniProfiler = (function() {
   };
 
   var bindDocumentEvents = function bindDocumentEvents() {
+    document.addEventListener("turbo:before-visit", e => {
```
kbrock, line 643: it looks like the other methods do not use anonymous blocks, they use functions defined. i.e.: onClickEvents, unbindDocumentEvents

Also, it looks like there is a check that `Turbolinks` is even relevant (by checking the constant defined in a gem. is there something similar you can do here?

Probably want to move this to the bottom of this method, right after Turbolinks section. That way the meat of this bindDocumentEvents method reads the same

lib/html/includes.js
```diff
@@ -640,6 +640,17 @@ var _MiniProfiler = (function() {
...
+      window.MiniProfilerContainer = document.querySelector('body > .profiler-results')
+      if(!e.defaultPrevented) window.MiniProfiler.pageTransition()
+    })
+
+    document.addEventListener("turbo:load", e => {
+      if(window.MiniProfilerContainer) {
```
kbrock, line 649: could you also enhance `unbindDocumentEvents` to remove these events?
I think that will remove a memory leak.

kbrock (commented):
This looks great. but please remember:

1. I am not a rack-mini-profiler developer. I do not have merge rights. (feel free to ignore my suggestions)
2. I am not a ui developer and just looking at the code and suggesting you copy stuff.

lib/html/includes.js
ceritium: Thanks, @kbrock. 

I refactored the callbacks as functions and moved the code under Turbolinks logic.

`Turbo` doesn't implement `supported` like `Turbolinks`, so I am just checking that `Turbo` is available.

Doing the Turbo constant check avoid adding unnecessary event listeners, but it requires that the application exposes `Turbo` globally:

```
import { Turbo } from "@hotwired/turbo-rails"
window.Turbo = Turbo
```

What do you think? What is the best approach? I guess we should mention it in the documentation.

lib/html/includes.js
ceritium: Done.

SamSaffron:
This looks right to me @OsamaSayegh should we merge?

OsamaSayegh (commented):
Thank you for this PR @ceritium ❤️ I'd love to merge this, but I have a couple of comments:

Re:

> Doing the Turbo constant check avoid adding unnecessary event listeners, but it requires that the application exposes Turbo globally:
>    ```
>    import { Turbo } from "@hotwired/turbo-rails"
>    window.Turbo = Turbo
>    ```
> What do you think? What is the best approach? I guess we should mention it in the documentation.

Requiring people to do something like that feels too much magical to me 😅 Instead, I'd prefer if we added a new `enable_hotwire_turbo_support` config option (false by default) to [`lib/mini_profiler/config.rb`](https://github.com/MiniProfiler/rack-mini-profiler/blob/a93060ef9d49c5e4007d00eb8bb5a783263bb220/lib/mini_profiler/config.rb) that people who use Hotwire in their application can turn on. When it's turned on, we can use that to signal to MiniProfiler to add Hotwrite's specific bits.

For example, in `lib/html/profile_handler.js` I'd replace the `data-turbo-permanent="true"` attribute that you have now with a new placeholder `{hotwriteTurboSupport}`. The `get_profile_script` method here:

https://github.com/MiniProfiler/rack-mini-profiler/blob/a93060ef9d49c5e4007d00eb8bb5a783263bb220/lib/mini_profiler/profiler.rb#L720-L774

is what replaces the placeholders in `profile_handler.js` with the right values. So, I'd add a check for our new config, and if it's true I'd set `hotwriteTurboSupport` in the `settings` hash to `data-turbo-permanent="true"`; otherwise set `hotwriteTurboSupport` to the empty string.

Finally, instead of checking for the `Turbo` constant on the client side, we can check for the `data-turbo-permanent` attribute on MiniProfiler's `<script>` tag and add our event listeners if the attribute exists.

The advantage of this is it allows to conditionally add the `data-turbo-permanent="true"` attribute only for the people who use Turbo, and avoid magical hacks like `window.Turbo = Turbo`.

One last thing, could run the `update_asset_version` rake task when you're done? This updates the `ASSET_VERSION` constant which use to bust cache when new versions are released.

Let me know if you have any questions!

ceritium:
Thanks for your feedback @OsamaSayegh. I like your approach. I just pushed a commit addressing your feedback and a small test. 

Would you add more tests?
Do you have any suggestions about variable naming? 
It's ok the place for `enable_hotwire_turbo_drive_support` on "config.rb" or would you move it to a different place?

OsamaSayegh:
I think it looks great, thank you @ceritium!

ceritium:
@OsamaSayegh I added the documentation. 

Do you like the commits as they are? Would you squash them? I didn't found any policy about that.

OsamaSayegh:
Perfect, thank you for all your work on this!

Github has a merge option that squashes everything together which I'm going to use now!

PR #500, opened 2021-06-15
Remove type attribute from includes script

bubba:
In HTML 5, the default since rails 3, the type attribute on scripts defaults to "text/javascript". Removing this fixes warnings being thrown by HTML 5 validators that complain "The type attribute is unnecessary for JavaScript resources", [specifically the W3 validator](https://validator.w3.org/nu/)

image


SamSaffron:
I am not sure about this, I support some opt-out I guess, but non HTML 5 still needs this.

bubba:
Understandable, wasn't sure if HTML4 was still supported or not

PR #501, opened 2021-06-15
Allow content-type header with charset on test.

ceritium:
Specs fails when the Content-Type is "text/plain; charset=utf-8"


SamSaffron:
Thanks heaps!

PR #502, opened 2021-06-20
Improve development section from Readme.md

ceritium:
- Add note about bundle install on `website/`
- Add note about required services for specs
- Remove autotest note. It was removed on https://github.com/MiniProfiler/rack-mini-profiler/pull/21


SamSaffron:
looks good, thanks!

PR #507, opened 2021-08-19
Add a note to the README about SQL timings absence when generating flamegraphs

### LINKED ISSUE #506: Flamegraph & % in sql queries don't show up
I'm seeing some odd behavior in the 2.3.2 release where:

- async-flamegraphs don't show any meaningful information (this might be expected, since this feature was added recently)
- The `% in sql` annotation doesn't show any SQL queries (this isn't expected, since it worked previously)

I think this is maybe a bug, or regression, so I've provided some information to try to describe what I expect to see, what I actually see, and why I expect what I expect (tldr; it's what I've seen in earlier versions of `rack-mini-profiler`)

I've got a demo repository & replication instructions here: https://github.com/jaydorsey/profiler

I _think_ this is a fairly vanilla setup, but I'm testing an API endpoint. Maybe I'm just using the tool incorrectly? Another set of eyes would be appreciated.

Examples of what rack-mini-profiler shows in the latest gem release:

No SQL queries (in any of the requests that show up):

image

They do show up in the server logs:

image

The flamegraphs have no meaningful data:

image

I've tried to `curl` the endpoint to confirm it loads fine; no server errors as far as I can tell.

Things I've tried:

- Using the `master` branch of the gem
- Setting `export RACK_MINI_PROFILER_PATCH="pg"`
- Regressing and/or removing some of my dependencies (bootsnap, listen, spring, pg)
- `bundle update` to make sure all my dependencies are updated
- Rails 6.1.3 and 6.1.4
- Ruby 3.0.1 and 3.0.2
- Regressing back to rack-mini-profiler:
  - 2.2.0, 2.2.1, 2.3.0, and 2.3.1 all "work"\*, they show the queries (didn't test flamegraph)
  - 2.3.2 doesn't work

This behavior happens in another repository as well (a real application that has many more dependencies)

I looked thru the diff between 2.3.1 and 2.3.2 and I don't see anything at all SQL related that might cause this. Another engineer confirmed the same behavior on their computer (to try to confirm it wasn't a local setting)

Was wondering if anyone else can confirm the behavior and/or maybe sees something that I missed?

One other thing I noted when looking at stackprof itself:

- When I use `wall` mode, like rack-mini-profiler does, I see similar results in what `stackprof` generates; there's no meaningful info
- When I use `cpu` mode with `stackprof`, I get the results that I expected `rack-mini-profiler` to return (you can see the rails calls in there)

image

The mode doesn't appear configurable in `rack-mini-profiler` so I'm not sure if this was intentional or if I'm just using the tool incorrectly.

\* No async-flamegraph; assuming that was added recently

OsamaSayegh:
Closes #506

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,9 @@ To store flamegraph data for later viewing, append the `?pp=async-flamegraph` pa
 Flamegraph data for this request, and all subsequent requests made by this page (based on the `REFERER` header) will be stored.
 'flamegraph' links will appear for these requests in the MiniProfiler UI.
 
+Note: Mini Profiler will not record SQL timings for a request if it asks for a flamegraph. The rationale behind this is to keep
+Mini Profiler's methods that are responsible for generating the timings data out of the flamegraph.
+
 ### Memory Profiling
 
 Memory allocations can be measured (using the [memory_profiler](https://github.com/SamSaffron/memory_profiler) gem)


smridge (approved):

smridge:
Thank you for this!

PR #516, opened 2022-01-25
Fix deprecated uses of `Redis#pipelined`

casperisfine:
Context: https://github.com/redis/redis-rb/pull/1059

The following is deprecated
```ruby
redis.pipelined do
  redis.get(key)
end
```

And should be rewritten as:
```ruby
redis.pipelined do |pipeline|
  pipeline.get(key)
end
```

Functionally it makes no difference.


OsamaSayegh (approved):

OsamaSayegh:
Thank you @casperisfine!

PR #517, opened 2022-01-27
[#479] Documentation update for SPA usage

### LINKED ISSUE #479: Setup for SPA not working
I ran into some issues\ trying to install the profiler on a Vue app. First, I had to modify the `script` tag from the example in the README, as it was missing two required data attributes (`data-hidden-custom-fields data-html-container="body"`).

Then I was getting this error in the console: `POST http://localhost:3000/mini-profiler-resources/results 404 (Not Found)`. Which I traced to using "wrong" `data-id` / `data-ids` values.

I ended up with this:

```html
<script type="text/javascript" id="mini-profiler"
        src="/mini-profiler-resources/includes.js?v=10da952c710f6abd9f1bd50fe50ed714"
        data-css-url="/mini-profiler-resources/includes.css?v=10da952c710f6abd9f1bd50fe50ed714"
        data-version="10da952c710f6abd9f1bd50fe50ed714" data-path="/mini-profiler-resources/"
        data-horizontal-position="left" data-vertical-position="top"
        data-ids=""
        data-trivial="false" data-children="false" data-max-traces="20" data-controls="false"
        data-total-sql-count="false" data-authorized="true" data-toggle-shortcut="alt+p" data-start-hidden="false"
        data-collapse-results="true" data-html-container="body" data-hidden-custom-fields></script>
```

I could create a PR for updating the docs if that helps (and above looks right).

katelovescode:
Updates the documentation to include @ivanyv 's fixed SPA script, as well as some clarity on using a conditional when MiniProfiler isn't in the window state yet, and a Gist for how to use it in React's useEffect hook.

Fixes #479

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,9 @@ Single page applications built using Ember, Angular or other frameworks need som
 On route transition always call:
 
 ```
-window.MiniProfiler.pageTransition();
+if (window.MiniProfiler !== undefined) {
+  window.MiniProfiler.pageTransition();
+}
 ```
 
 This method will remove profiling information that was related to previous page and clear aggregate statistics.
--- a/README.md
+++ b/README.md
@@ -353,13 +353,33 @@ window.MiniProfiler.pageTransition();
 This method will remove profiling information that was related to previous page and clear aggregate statistics.
 
 #### MiniProfiler's speed badge on pages that are not generated via Rails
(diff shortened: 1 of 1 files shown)


SamSaffron:
Thanks Kate, appreciate the improvement to the documentation!

PR #519, opened 2022-02-07
Improve GitHub Actions workflows

mishina2228:
* Enable `bundler-cache` in ruby/setup-ruby
* Disable `fail-fast` not to cancel in-progress jobs if any matrix job fails
  * This will improve the situation where recent CIs failed in Ruby 2.6 and other matrices are not running.


SamSaffron:
Nice changes, yes, happy to include them to assist in debugging the flaky build!

PR #520, opened 2022-02-09
Fix ArgumentError: comparison of String with 200 failed

mickeytgl:
Casts the `status` to an integer before comparison.

### Issue
* #521

### Related PRs and Issues
* Issue #185 
* PR #186

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -391,6 +391,7 @@ module Rack
           status, headers, body = [200, { 'Content-Type' => 'text/html' }, [blank_page_html.dup]]
         else
           status, headers, body = @app.call(env)
+          status = status.to_i
         end
       ensure
         trace.disable if trace


SamSaffron:
I am very mixed on this ... who up in the stack is sending a "200" string? I accepted this in memory_profiler years ago, but there have been so many intervening years to fix this and something up your stack is returning strings as status which is not correct.

mickeytgl:
👋🏻  Thanks for the quick response, yeah, I definitely agree that the true fix would be to dig into my stack call to see what is generating a 200 string, and I am trying to find out the culprit there. That being said, I think it's also nice to have this small safety net and that it does not add too much overhead having it, but of course, it's just a proposal

mickeytgl:
Hello 👋🏻  Bumping this to either get it in, further discuss it or close it

SamSaffron:
Yeah I prefer this be fixed upstream, feels odd to be responsible for fixing bugs further up the middleware stack.

PR #524, opened 2022-02-17
FEATURE: add cookie path support

featheredtoast:
Adds cookie path support for sites hosting cookies on subfolder paths.

Without this, other set-cookie headers' paths may be unable to set paths.

### DIFF
--- a/lib/mini_profiler/config.rb
+++ b/lib/mini_profiler/config.rb
@@ -17,6 +17,7 @@ module Rack
         new.instance_eval {
           @auto_inject      = true # automatically inject on every html page
           @base_url_path    = "/mini-profiler-resources/".dup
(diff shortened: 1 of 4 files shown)


lib/mini_profiler/config.rb
```diff
@@ -17,6 +17,7 @@ def self.default
         new.instance_eval {
           @auto_inject      = true # automatically inject on every html page
           @base_url_path    = "/mini-profiler-resources/".dup
+          @cookie_path      = "/".dup
```
SamSaffron, line 20: these `.dup` calls sure are confusing...

SamSaffron (approved):

SamSaffron:
looks good, we can nuke the .dup calls later.

Issue #528, opened 2022-02-25
No longer compatible with Ruby 2.4


DRBragg:
Due to changes in #518 `rack-mini-profiler is no longer compatible with Ruby 2.4. When attempting to use it in a Rails 5.0 and Ruby 2.4 environment attempting to boot the rails server will result in the following error:
<details closed>
<summary>
/var/bigred/shared/bundler/ruby/2.4.0/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `require': 

/var/bigred/shared/bundler/ruby/2.4.0/gems/rack-mini-profiler-3.0.0/lib/mini_profiler/storage/redis_store.rb:238: syntax error,

 unexpected keyword_rescue, expecting keyword_end (SyntaxError)
        rescue
              ^
/var/bigred/shared/bundler/ruby/2.4.0/gems/rack-mini-profiler-3.0.0/lib/mini_profiler/storage/redis_store.rb:352: syntax error, 
unexpected keyword_end, expecting end-of-input

</summary>
	from /var/bigred/shared/bundler/ruby/2.4.0/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `block in require'
	from /var/bigred/shared/bundler/ruby/2.4.0/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:259:in `load_dependency'
	from /var/bigred/shared/bundler/ruby/2.4.0/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `require'
	from /var/bigred/shared/bundler/ruby/2.4.0/gems/rack-mini-profiler-3.0.0/lib/rack-mini-profiler.rb:21:in `<top (required)>'
	from /usr/lib/rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:81:in `require'
	from /usr/lib/rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:81:in `block (2 levels) in require'
	from /usr/lib/rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:76:in `each'
[...]

PR #529, opened 2022-02-25
Update gem and docs to ref Ruby 2.5

DRBragg:
As noted in #528 `rack-mini-profiler` is no longer compatible with Ruby 2.4. This PR simply updates the README and gemspec to reference that

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'http://rubygems.org'
-ruby '>= 2.4.0'
+ruby '>= 2.5.0'
 
 gemspec
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'http://rubygems.org'
-ruby '>= 2.5.0'
+ruby '>= 2.6.0'
 
 gemspec
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ If you feel like taking on any of this start an issue and update us on your prog
 
 ## Installation
 
-Install/add to Gemfile in Ruby 2.4+
+Install/add to Gemfile in Ruby 2.5+
 
 ```ruby
 gem 'rack-mini-profiler'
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'http://rubygems.org'
-ruby '>= 2.4.0'
+ruby '>= 2.5.0'
```
OsamaSayegh, line 4:
```suggestion
ruby '>= 2.6.0'
```

README.md
```diff
@@ -27,7 +27,7 @@ If you feel like taking on any of this start an issue and update us on your prog
 
 ## Installation
 
-Install/add to Gemfile in Ruby 2.4+
+Install/add to Gemfile in Ruby 2.5+
```
OsamaSayegh, line 30:
```suggestion
Install/add to Gemfile in Ruby 2.6+
```

rack-mini-profiler.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
     "CHANGELOG.md"
   ]
   s.add_runtime_dependency 'rack', '>= 1.2.0'
-  s.required_ruby_version = '>= 2.4.0'
+  s.required_ruby_version = '>= 2.5.0'
```
OsamaSayegh, line 24:
```suggestion
  s.required_ruby_version = '>= 2.6.0'
```

OsamaSayegh (commented):
Thank you for this! We don't intentionally break support for EOL ruby versions, but we make no effort to support them either. 2.5 is also EOL so let's bump the minimum to 2.6.

DRBragg:
> Thank you for this! We don't intentionally break support for EOL ruby versions, but we make no effort to support them either. 2.5 is also EOL so let's bump the minimum to 2.6.

Makes sense to me! Thanks for the quick response!

OsamaSayegh:
Thank you for contribution @DRBragg!

PR #532, opened 2022-03-01
Use Error#original_message for RUBY_VERSION >= 3.1

muellerj:
Related rspec issue: rspec/rspec-expectations#1338

### DIFF
--- a/lib/mini_profiler/profiler.rb
+++ b/lib/mini_profiler/profiler.rb
@@ -538,8 +538,7 @@ module Rack
 
         body << "\nBacktraces\n"
         exceptions.each_with_index do |e, i|
-          message = Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.1") ? e.message : e.message.lines.first.chomp
-          body << "##{i + 1}: #{e.class} - \"#{message}\"\n  #{e.backtrace.join("\n  ")}\n\n"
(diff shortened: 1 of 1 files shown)


lib/mini_profiler/profiler.rb
```diff
@@ -538,7 +538,8 @@ def dump_exceptions(exceptions)
 
         body << "\nBacktraces\n"
         exceptions.each_with_index do |e, i|
-          body << "##{i + 1}: #{e.class} - \"#{e.message}\"\n  #{e.backtrace.join("\n  ")}\n\n"
+          message = Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.1") ? e.message : e.message.lines.first.chomp
+          body << "##{i + 1}: #{e.class} - \"#{message}\"\n  #{e.backtrace.join("\n  ")}\n\n"
```
OsamaSayegh, line 542: Thank you for this! I noticed the tests were failing on ruby 3.1 and I meant to debug it but got distracted with other things!

I wonder if we should just take the first line of the error message unconditionally? Because the way exceptions are formatted is that each exception gets a "header" line that contains the exception number, class and message and then the header line is followed by the backtrace. If we get an exception where the message is multi-line, then the header for that exception becomes 2 lines and it looks weird.

OsamaSayegh:
Perfect, thank you @muellerj!

PR #538, opened 2022-09-04
Remove always use prepend instead of alias_method for patches

johnnyshields:
There are two patches which have options to use prepend instead of alias_method (alias_method is currently used by default):

- prepend_net_http_patch
- prepend_mysql2_patch

It would be much better if Rack Mini Profiler always uses prepend for these, as there is no disadvantage to doing so. I spent 3 hours putzing around trying to figure out why Rack Mini Profiler wasn't working, and I discovered I had to include the Net::HTTP prepend patch (which Rack Mini Profiler could have just done by default.)


OsamaSayegh:
@johnnyshields I wish we could remove the options and just use prepend to apply our patches, but unfortunately I don't think we can because if another gem is installed alongside MP and happens to also patch a method that MP patches using `alias_method`, then that method will error when it's called. See [this issue](https://bugs.ruby-lang.org/issues/11120) on ruby's bug tracker.

johnnyshields:
@OsamaSayegh maybe `prepend` should be the default then, and we have a patch to enable alias instead? Rails for example removed usage of `alias_method_chain` in favor of `prepend`, so most other gems *should* use prepend if not doing so already.

nateberkopec:
I've dealt with this one personally and I think it's not worth the effort to change the default here. Maybe someday in an alias_method-less world 5 years from now (someone bookmark this in 5 years and ping me, you legend).

PR #539, opened 2022-09-04
Remove Rails 2.x initialization from Readme

johnnyshields:
Rack Mini Profiler requires Ruby 2.6, which can't be used with Rails 2.x


nateberkopec (approved):

SamSaffron:
thanks!

PR #543, opened 2022-12-26
Adds Ruby 3.2 to the CI matrix

petergoldstein:
Updates checkout action version.

I also needed to change a `127.0.0.1` to `localhost` to get the specs to pass.

PR #547, opened 2023-02-07
Fix stackprof not installed error

gmcgibbon:
Fixes malformed response when `pp=flamegraph` is used and stackprof is not installed (this happens on web servers such as puma due to the expectation that the body is an array, it crashes here: https://github.com/puma/puma/blob/d8765e362dd9b273635facefa0911252abdb584b/lib/puma/request.rb#L153).

```
2023-02-07 15:15:34 -0600 Read: #<NoMethodError: undefined method `each' for "Please install the stackprof gem and require it: add gem 'stackprof' to your Gemfile":String>
```

Also forwards headers and status code from app response error cases so that app logs are consistent with what is actually returned by middleware.

I couldn't test this through integration tests, so I had to use units. If we don't see value in the well-formed response tests, I can remove them.

### DIFF
--- a/spec/lib/profiler_spec.rb
+++ b/spec/lib/profiler_spec.rb
@@ -198,4 +198,35 @@ describe Rack::MiniProfiler do
       expect(Rack::MiniProfiler.snapshots_transporter?).to eq(true)
     end
   end
+
+  describe '#call' do
+    let(:app) { lambda { |env| [200, {}, ["OK"]] } }
+    let(:profiler) { Rack::MiniProfiler.new(app) }
+
+    it "returns error response when stackprof isn't installed" do
+      response = profiler.call({ "PATH_INFO" => "/", "QUERY_STRING" => "pp=flamegraph" })
+
+      expect(response).to eq([
+        200,
+        { "Content-Type" => "text/plain; charset=utf-8", "Set-Cookie"=>"__profilin=p%3Dt; path=/; HttpOnly; SameSite=Lax" },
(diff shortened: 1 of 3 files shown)


gmcgibbon:
Looks like CI errors are due to change in behaviour from the redis client. There's a fix on https://github.com/MiniProfiler/rack-mini-profiler/pull/543 to get CI green.

spec/lib/profiler_spec.rb
```diff
@@ -198,4 +198,27 @@ def self.bar(baz, boo)
...
+    let(:app) { lambda { |env| [200, {}, ["OK"]] } }
+    let(:profiler) { Rack::MiniProfiler.new(app) }
+
+    it "returns well-formed response when stackprof isn't installed" do
+      response = profiler.call({ "PATH_INFO" => "/", "QUERY_STRING" => "pp=flamegraph" })
+      status, headers, body = response
+
+      expect(100..599).to include(status)
+      expect(headers.keys).to(all(be_a(String)))
+      expect(body).to(all(be_a(String)))
```
nateberkopec, line 212: Rather than check for an Enumerable of Strings, lets do the Rack-spec thing and check for response to each?

nateberkopec (changes requested):
Spec change and I think it's G2G

spec/lib/profiler_spec.rb
gmcgibbon: Done. This caught a bug in the memory profiler spec, so thanks for that!

nateberkopec:
Needs a rebase and then we're set.

PR #548, opened 2023-02-07
Require Rails 6.0+

gmcgibbon:
Use bundler to require Rails 6.0 or greater as a peer dependency. We should use Rails 6+ here because that is what's currently supported in the Ruby on Rails maintenance policy.

Also looks like https://github.com/MiniProfiler/rack-mini-profiler/pull/539 can be merged with this change. Rails 2.x support will be dropped. If a developer wants to reference this for an older version of the profiler, they can switch to the version tag.

For reference, this approach is inspired by how we manage peer deps in Rails: https://github.com/rails/rails/blob/dcd8f37bc63d46f28aa55e6891b03a0b0b73e497/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L6

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+gem "rails", ">= 6.0"
+
 require 'fileutils'
 require_relative './railtie_methods'
 
@@ -154,13 +156,7 @@ module Rack::MiniProfilerRails
       return false
     end
 
-    if ::Rails.version >= "5.0.0"
-      ::Rails.configuration.public_file_server.enabled
-    elsif ::Rails.version >= "4.2.0"
-      ::Rails.configuration.serve_static_files
-    else
-      ::Rails.configuration.serve_static_assets
-    end
+    ::Rails.configuration.public_file_server.enabled
   end
 
   class Railtie < ::Rails::Railtie
--- a/lib/rack-mini-profiler.rb
+++ b/lib/rack-mini-profiler.rb
@@ -32,6 +32,4 @@ require 'mini_profiler/profiler'
 require 'patches/sql_patches'
 require 'patches/net_patches'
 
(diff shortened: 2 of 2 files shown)


nateberkopec:
I'm usually not in favor of changing version requirements based on EOL policies, but rather on actual support for the versions in question.

Why should we drop support for old Rails versions that still work? RMP is really the only performance tool available for old Rails applications, why should we lock those apps out of newer versions of RMP?

nateberkopec (commented):
See comment

gmcgibbon:
> I'm usually not in favor of changing version requirements based on EOL policies, but rather on actual support for the versions in question.

That's fair, I agree under most circumstances we should only drop support for Ruby/Rails versions when we have a real need to do so (via a code change etc.).


> Why should we drop support for old Rails versions that still work? RMP is really the only performance tool available for old Rails applications, why should we lock those apps out of newer versions of RMP?

I think what I was trying to clarify with Rack Mini Profiler is that it should support _some_ minimum version of Rails. Right now, we seem to specify a minimum of 3.0 (because the Railtie doesn't get included below that). We also don't have any CI integration tests to make sure it works with _any_ Rails version, so I think it is unclear what this gem supports and what it doesn't. Clarifying expectations here would be helpful for maintainers of the library, and for uses of the library.

As an aside, Rails developers who are still using Rails <6 don't have security patch support, and are running vulnerable version web applications that need to be updated anyway. To me, this is just more incentive to upgrade, but I don't feel strongly about this. Happy to close, or lower the version requirement. Let me know what you think.

nateberkopec:
> Right now, we seem to specify a minimum of 3.0 (because the Railtie doesn't get included below that).

I think it makes sense to merge that hard requirement at 3.0 today (because as you say, we literally know the gem doesnt work for 2.0), and we can look at adding specs in the future. If the specs tell us our support for 3.0/4.0 etc is broken, we can make the decision at that point to change Rails version support in the gemspec.

Issue #349, opened 2018-05-17
I'm using Mongoid but getting Gem Load Error is: uninitialized constant ActiveRecord::ConnectionAdapters


rgaufman:
Hi there, I'm using rails (5.2.0), rack-mini-profiler (1.0.0) with mongoid (6.4.1) - I don't have (or want) ActiveRecord.

When I run rails c, I get this:

```
✗ rails c
Traceback (most recent call last):
	18: from bin/rails:4:in `<main>'
	17: from bin/rails:4:in `require'
	16: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/commands.rb:18:in `<top (required)>'
	15: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/command.rb:46:in `invoke'
	14: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/command/base.rb:65:in `perform'
	13: from /usr/local/lib/ruby/gems/2.5.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
	12: from /usr/local/lib/ruby/gems/2.5.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
	11: from /usr/local/lib/ruby/gems/2.5.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
	10: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/commands/console/console_command.rb:95:in `perform'
	 9: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/command/actions.rb:17:in `require_application_and_environment!'
	 8: from /usr/local/lib/ruby/gems/2.5.0/gems/railties-5.2.0/lib/rails/command/actions.rb:17:in `require'
	 7: from /Users/hackeron/Dropbox/Development/Xanview/timeline/config/application.rb:26:in `<top (required)>'
	 6: from /usr/local/Cellar/ruby/2.5.1/lib/ruby/2.5.0/bundler.rb:114:in `require'
	 5: from /usr/local/Cellar/ruby/2.5.1/lib/ruby/2.5.0/bundler/runtime.rb:65:in `require'
	 4: from /usr/local/Cellar/ruby/2.5.1/lib/ruby/2.5.0/bundler/runtime.rb:65:in `each'
[...]

PR #549, opened 2023-02-08
Make patches opt-in

gmcgibbon:
This might be a contentious change. For developers that only use rack-mini-profiler for call stack and/or memory profiling, speed badge granularity doesn't matter as much. In fact, it can add extra stack frames where we would expect there not to be. Hence, they shouldn't need to be required to patch their SQL and HTTP clients in their applications. 

So, I think deprecating `SqlPatches.patch` and making patches opt-in (manually placed in an initializer or in `application.rb` of their Rails app or equivalent for other frameworks) is a good solution for those who don't want this instrumentation forcibly patched into their apps.

If this is accepted, I would be interested in making speed badge injection configurable. Though, this might defeat the purpose of the library if we expect the developer to always want speed badges. Let me know what you think. My view of the library may be different from yours, so I'm happy to discuss and add/remove/edit commits.

While I was here, I also:
- Cleaned up the load path so everything expect `sql_patches` (which I would like to see removed in the next version of this gem) so that everything is under `mini_profiler`.
- Require patched libraries before patching them so the patch will always work. I believe this should be safer if we assume that patches are opt-in and required when the app developer expects them to be defined.
- Split `SqlPatches` into a module that actually patches, and one that patches call into.


nateberkopec (changes requested):
I'm a "No" on this change. 

Removing Net::HTTP and SQL instrumentation by default makes this require too much effort to install, because 80% of people at least will want these patches on.

However, I think it's a great idea to make it easier to run RMP in "patchless" or "100% opt-in" mode. I'm not a huge fan of the current ENV variable based way. I think we could probably have a "patchless" entrypoint that requires only RMP without patches, and the main/default entry point requires that entrypoint + all the patches we currently require. WDYT?

gmcgibbon:
> I think we could probably have a "patchless" entrypoint that requires only RMP without patches, and the main/default entry point requires that entrypoint + all the patches we currently require. WDYT?

That seems fair. I would prefer patches to be turned off by default, but this is mainly due to my usage objectives. I also understand the usefulness of the patches. I'll get to work on refactoring this shortly, thanks!

nateberkopec:
Closes #349 #349

nateberkopec:
It looks like the main thing that would have to be opted-out would be the Net::HTTP patch. I had hoped to extend `RACK_MINI_PROFILER_PATCH` to support this but it would be a breaking change, as `RACK_MINI_PROFILER_PATCH=false` would now stop patching Net::HTTP.

An intermediate interface, which we could clean up later, would be to just add a new env variable like `RACK_MINI_PROFILER_PATCH_NET_HTTP`.

PR #550, opened 2023-02-08
Deprecate rack_profiler:install in favour of rack_mini_profiler:install

gmcgibbon:
The gem's name and install generator should be the same.

This deprecates the current generator in favour of a more appropriately named one. It also updates copy on the initializer comments and usage clarifying what it is doing and when to use an initializer. Behaviour is changed for the install generator to always generate a `rack_mini_profiler.rb` initializer instead, but I think that's acceptable if we want to rename the generator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ be loaded outright, and an attempt to re-initialize it manually will raise an ex
 Then run the generator which will set up rack-mini-profiler in development:
 
 ```bash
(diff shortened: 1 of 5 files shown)


nateberkopec (approved):

SamSaffron:
I think the reason I was holding off on merging stuff was cause our test suite is not passing, looks unrelated though.

PR #551, opened 2023-02-08
Simplify requires

gmcgibbon:
- Fix file naming for Rack::MiniProfiler
- Make module files require their nested contents
- Move library requires to the files that use them
- Remove require for timeout because it isn't used
- Remove require for thread because it is required by default

This patch also allows for easier transition to autoloading should we
choose to use it later.


nateberkopec:
Rebase on main for the latest CI and I think this LGTM!

PR #552, opened 2023-02-08
Add titles to middleware pages

gmcgibbon:
Right now the pages you visit that route to the middleware don't have HTML titles or valid markup. This fixes that.


nateberkopec:
Rebase and LGTM 👍

PR #553, opened 2023-02-08
Profile parameter config

gmcgibbon:
Allow profile query parameter to be configured using `Rack::MiniProfiler.profile_parameter`. Sometimes, an application may want to use a configurable profiler parameter to obscure or customize how profiles are collected. This is especially true for production environments.

I also formatted the config option table to make it easier to read without markdown parsing, but I can revert that if need be. It really only helps if you have a 120 character screen width.


nateberkopec:
It does make the code a tad harder to read but I definitely think this should be configurable, so LGTM. Rebase to show you're green on CI and let's 🚢

PR #555, opened 2023-02-08
Autoload sotrage adapters

gmcgibbon:
Autoloads storage adapters because it is unnecessary to load storage
adapters you won't use. Autoload lets us defer loading the file until it
is referenced, so that we can require storage dependencies normally and
expect them to load when called.

Depends on https://github.com/MiniProfiler/rack-mini-profiler/pull/551. I can't make a PR to this repo if my merge target is on my fork, so this will have to do for now.

### DIFF
--- a/lib/mini_profiler.rb
+++ b/lib/mini_profiler.rb
@@ -27,11 +27,11 @@ module Rack
       end
 
       def resources_root
-        @resources_root ||= ::File.expand_path("../../html", __FILE__)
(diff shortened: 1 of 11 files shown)


suryanarayanan035:
The code changes Looks good to me. We just need to sync the **autoload_storage_adapaters** with the latest master  and resolve the conflict. After @gmcgibbon is done with the mentioned action items, we can merge this code @nateberkopec .

PR #556, opened 2023-03-25
Prepend patch for PG

okarsono:
Similar to the mysql2 patch, this workaround is for PG (Postgresql) users who are experiencing conflict between `rack-mini-profiler` and another gem that does the same patching.


rqz13:
I'd love to see this merged.

nateberkopec:
I kinda screwed this up, I should've done this manually on the command line and added @okarsono as co-author but I did this instead. Sorry!

Issue #561, opened 2023-03-28
Rails controller actions should display as ControllerName#action


nateberkopec:
See #391

keshavbiswa:
This issue is fixed. @nateberkopec Can we close this?

suryanarayanan035:
I see that this PR is merged but there is no test cases written. I think we can close this issue if test cases are not needed else let us write test cases. your thoughts on this @nateberkopec ?

PR #567, opened 2023-04-05
Add controller_name to Description

keshavbiswa:
This is an enhancement based on #561 .

Controller without namespace will now render the controller name:

Screenshot 2023-04-05 at 10 23 53 PM

Controller with namespaces will render the namespaced controller name:

Screenshot 2023-04-05 at 10 24 32 PM

PS: Need help with the tests!

cc: @nateberkopec

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -75,7 +75,8 @@ module Rack::MiniProfilerRails
         next if !should_measure?
 
         current = Rack::MiniProfiler.current
-        description = "Executing action: #{payload[:action]}"
+        controller_name = payload[:controller].sub(/Controller\z/, '').downcase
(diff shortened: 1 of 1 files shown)


lib/mini_profiler_rails/railtie.rb
```diff
@@ -75,7 +75,8 @@ def self.initialize!(app)
         next if !should_measure?
 
         current = Rack::MiniProfiler.current
-        description = "Executing action: #{payload[:action]}"
+        controller_name = payload[:controller].sub(/Controller\z/, '').downcase
```
nate-at-gusto, line 78: I'm not sure if I like the downcasing or not, but that's easy enough to change later.

nate-at-gusto:
Test suite is a bit of a mess, punting it for now

PR #584, opened 2023-06-11
Use profile parameter config for framegraph ability

ohbarye:
Profile parameter got config-able with https://github.com/MiniProfiler/rack-mini-profiler/pull/553, but the framegraph ability still depends on `pp`. I think it should use `@config.profile_parameter`.


nateberkopec:
Great catch! Fixed by fb2c080de919db564f6d2f2f39cdb842f3a65c7a on master now.

PR #587, opened 2023-07-03
Include configured base path in speedscope iframe URL

davidtaylorhq:
When rack-mini-profiler was used under a different base URL (e.g. when deployed in a Rails app with `relative_url_root` configured), the stackprof iframe URL was missing the base path. This commit fixes the path for the flamegraph by sharing logic with the `#get_profile_script` method, and adds a spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


tgxworld (approved):

OsamaSayegh (approved):
LGTM, thanks!

PR #591, opened 2023-10-18
Check if memcached is running during memcache spec, and fail if not running

technicalpickles:
I was trying to run the test suite locally, and it failed here because I didn't have memcached running (yet):

```
  1) Rack::MiniProfiler::MemcacheStore page struct storage can store a PageStruct and retrieve it
     Failure/Error: @client.set("#{@prefix}#{page_struct[:id]}", Marshal::dump(page_struct), @expires_in_seconds)

     Dalli::RingError:
       No server available
     # ./lib/mini_profiler/storage/memcache_store.rb:22:in `save'
     # ./spec/lib/storage/memcache_store_spec.rb:17:in `block (4 levels) in <top (required)>'
```

This adds a `alive?` method that the spec can check, and give more specific information to the user. I've also included a docker-compose.yml to give a one-liner to start memcached.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,6 @@
+# convenient way to start memcached for specs that need it: docker-compose up -d
+services:
+  memcached:
+    image: memcached
+    ports:
+      - "11211:11211"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,0 @@
-# convenient way to start memcached for specs that need it: docker-compose up -d
-services:
-  memcached:
-    image: memcached
-    ports:
-      - "11211:11211"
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+# convenient way to start memcached for specs that need it: docker-compose up -d
+services:
```
nateberkopec, line 2: I'm not yet ready to say this repo needs this file

nateberkopec (changes requested):

docker-compose.yml
technicalpickles: Fair. There isn't much to it, but does give a simple way to manage it. Removed for now and suggest a docker command instead.

PR #592, opened 2023-10-18
Refactor initial skip logic into methods in MiniProfiler

technicalpickles:
@nateberkopec mentioned in passing how big the `call` method is. This is just my attempt to reduce it a little 😁


technicalpickles:
This PR kinda got away from me 😁 It ended up being more about abstracting what settings are being passed in via query parameters than the skip logic.

I made a QuerySettings which encapsulates the logic. I think this has nice parity with ClientSettings, because we have places that end up like `query_settings.something || query_settings.something` which feels nice.

As I was going through this, I saw that there are a bunch of places doing `=~` to check matches, but not using the results. I changed those to `match?` to avoid allocating match data.

nateberkopec (approved):
This is a big improvement for readability. 👍

technicalpickles:
Have some merge conflicts to sort out in the wake of https://github.com/MiniProfiler/rack-mini-profiler/pull/593

SamSaffron:
I like this change a lot but yeah we need a rebase here.

kbrock:
Darn. This may be tricky to merge.

I had done something like this years back.
If it isn't merged right away, it easily becomes un-rebase-able.
(especially with 17 commits)

Thanks for this PR Josh. (and ✋ )

PR #593, opened 2023-10-18
Refactor html and page serveing logic into modules

technicalpickles:
Another attempt to try to cleanup MiniProfiler. I started trying to group similar logic together to pull out to its own module, and the ones that stood out had to do with serving an HTTP response directly (ie actions) and generating HTML for said response.

That lead me to introducing `Actions` and `HTML` modules. This is just grouping methods into logical modules and including them, so it really change much functionally. It does let you see similar methods a lot closer together.

For `Actions`, most of the methods were already named `serve_`, so I moved those to Actions. Anything else that was serving up a request directly was moved there and named as well.

I changed `serve_html` a little because despite the name, it does serve up other things. I hoisted those conditionals for snapshot, results, and flamegraphs out of the method to the one place where serve_html was called.


lib/mini_profiler.rb
```diff
@@ -262,7 +195,15 @@ def call(env)
       end
 
       # handle all /mini-profiler requests here
-      return client_settings.handle_cookie(serve_html(env)) if path.start_with? @config.base_url_path
+      if path.start_with? @config.base_url_path
+        file_name = path.sub(@config.base_url_path, '')
+
+        return serve_results(env) if file_name.eql?('results')
```
nateberkopec, line 201: `case`? 🤔

lib/mini_profiler.rb
```diff
@@ -279,46 +220,23 @@ def call(env)
...
 
+      # remember that profiling is not disabled (ie enabled)
+      client_settings.disable_profiling = false
+
       # profile gc
       if query_string =~ /#{@config.profile_parameter}=profile-gc/
-        return tool_disabled_message(client_settings) if !advanced_debugging_enabled?
-        current.measure = false if current
-        return client_settings.handle_cookie(Rack::MiniProfiler::GCProfiler.new.profile_gc(@app, env))
+        return serve_profile_gc(env, client_settings)
```
nateberkopec, line 230: I think we should keep `current.measure = false` up at this level, it's not a view level concern.

lib/mini_profiler/actions.rb
```diff
@@ -0,0 +1,153 @@
+module Rack
+  class MiniProfiler
+    module Actions
+      def serve_snapshot(env)
+        self.current = nil
```
nateberkopec, line 5: Maybe should be top level as well.

nateberkopec (approved):
My comments are all nits, so marking as approve.

website/sample.rb
```diff
@@ -94,6 +94,7 @@ def create_fake_snapshot(method, path, duration, started_at)
   end
 end
 
+Rack::MiniProfiler.config.enable_advanced_debugging_tools = true
```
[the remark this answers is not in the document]
technicalpickles: Sorta. I didn't intend to commit this, but I needed to add it to be able to test everything. I can extract it to another PR though.

lib/mini_profiler.rb
technicalpickles: Yeah, good call. This was the code in the original method, but can change it while we are here

lib/mini_profiler.rb
technicalpickles: That makes sense, but it is slightly different logic. It's turning off measure _after_ serving the `tool_disabled_message`. It might be fine though, can see how tests like it.

Issue #514, opened 2021-12-17
Feature request: Ability to pass ignore_gc to StackProf.run


danny-pflughoeft:
In memory-heavy apps, the GC frames make the flamegraph basically unusable

<img width="1792" alt="Screen Shot 2021-12-16 at 5 27 05 PM" src="https://user-images.githubusercontent.com/90438738/146546317-9fd3d107-a83d-4072-96b1-afebf315ae22.png">

StackProf has an option to fix this: [`ignore_gc`](https://github.com/tmm1/stackprof#all-options). However it's not exposed by rack-mini-profiler

taka-ebato:
As a workaround, I add this code to a file in config/initializers. So far it's working fine.

```rb
module Rack
  class MiniProfiler
    class StackProf
      def self.run(mode:, raw:, aggregate:, interval:)
        ::StackProf.run(
          mode: mode,
          raw: raw,
          aggregate: aggregate,
          interval: interval,
          ignore_gc: true
        ) do
          yield
        end
      end
    end
  end
end
```

PR #599, opened 2023-12-06
Add option to ignore garbage collection frames with flamegraphs

ihollander:
Fixes https://github.com/MiniProfiler/rack-mini-profiler/issues/514

This adds a configuration option to ignore garbage collection via the [`ignore_gc`](https://github.com/tmm1/stackprof#all-options) option in StackProf. Users can override this configuration on a per-request basis via the `flamegraph_ignore_gc` query param.

I tested this change out on a separate app using this branch and confirmed garbage collection frames were visible by default, and hidden after setting `Rack::MiniProfiler.config.flamegraph_ignore_gc = true` or using the query param.


nateberkopec:
Something I've wanted for a long time! I know I just did a release but this is too useful to let set on main for a long time. I'll just do another minor version.

PR #601, opened 2023-12-09
Add flamegraph path to response headers

soberstadt:
The motivation for this is to enable easy access to the flamegraph when using the async-flamegraph action. Currently, if you have a CURL request that you want to profile, you have to add the `pp=async-flamegraph` query parameter to the request, then copy the first ID from the `X-MiniProfiler-Ids` header, then add that ID to the `/mini-profiler-resources/flamegraph?id=` path in the browser. This assumes that you know that the current request profile id is the first one in the `X-MiniProfiler-Ids` header (personally, I was not confident of this until I read the source code) and that you know the path to the flamegraph endpoint.

I originally added the protocol, host, and port to the path, but I didn't know if that was guaranteed to be correct in all cases, so I removed it. If anyone else has a preference for this, I would be happy to hear it!


SamSaffron:
this does feel right to me! thanks

PR #618, opened 2024-07-10
Fix redis_store.rb fetch_snapshots_group

dbachinin:


olleolleolle:
I think this can be closed, the existing code does the right thing.

kbrock:
Think ruby 2.5 introduced the ability to `rescue` in `do ; end` block.
Think this can be closed as well.

PR #621, opened 2024-08-20
Update README.md - bundle exec rails g rack_profiler:install

equivalent:
current README had a typo 

   bundle exec rails g rack_mini_profiler:install 

fails

    bundle exec rails g rack_profiler:install

works


coalest:
@equivalent Is it possible you were using an older version of the gem? Because `rack_profiler:install` was deprecated in 1d2f012c (v3.1.0).

SamSaffron:
this does not seem valid to me, sorry.

PR #623, opened 2024-10-01
respect flamegraph_mode in query string

onyxblade:
Currently, parameters like `?flamegraph_mode=cpu` are not correctly passed to StackProf. Instead, `config.flamegraph_mode` will always be used.

This PR fixes the parameter check and extraction.


tgxworld:
@onyxblade Thank you for your contribution. I have merged the changes in https://github.com/MiniProfiler/rack-mini-profiler/pull/635.

PR #624, opened 2024-10-07
Reset total SQL count on route transition

lucas-telus:
Fix the total SQL count not resetting on route transitions in SPAs when `show_total_sql_count` is enabled.

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -1179,6 +1179,7 @@ var _MiniProfiler = (function() {
 
       reqs = 0;
       totalTime = 0;
+      totalSqlCount = 0;
       expandedResults = false;
       toArray(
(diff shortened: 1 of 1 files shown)


coalest:
Test this using a simple rails app that sends a random (1-10) SQL requests for each request. (also with my patch for Turbo Drive).

Before this patch:
[Screencast from 2025-02-07 15-51-29.webm](https://github.com/user-attachments/assets/b24597f0-f915-4604-9ef0-b4dc5a38b922)

After this patch:
[Screencast from 2025-02-07 15-54-40.webm](https://github.com/user-attachments/assets/6470fbda-8dad-4092-a87e-8148a9b2ebb7)

LGTM

PR #629, opened 2025-01-30
Profiler totals row should not be part of results.

DropsOfSerenity:
The previous code had the chance of removing the profiler-totals when the max traces to show limit is hit, due to the profiler-results with profiler-totals row being the first row.

This would result on the rack mini profiler being unclickable and completely hidden when there are more than max traces to show. Breaking it entirely.

I think theres a bigger discussion about if we can support a way to see all traces, or paginate them in some way. But I think that's for another PR/discussion.


coalest:
Tested this out and it looks good to me!

I set the `max_traces_to_show` to be 5 so that the bug would be more apparent.

Before: (badge disappears after fifth trace added)

[Screencast from 2025-02-10 15-44-36.webm](https://github.com/user-attachments/assets/78b6bc0a-148c-49b4-87d0-1bda844e9a13)

After: (badge stays on the page)

[Screencast from 2025-02-10 15-45-57.webm](https://github.com/user-attachments/assets/c15c00b4-7b7a-47c9-a407-8fc4df15fcdf)

nateberkopec:
Thanks! Changelog entry please 🙏

DropsOfSerenity:
> Thanks! Changelog entry please 🙏

@nateberkopec Great! Will add it this week.

DropsOfSerenity:
@nateberkopec Added a changelog entry

PR #630, opened 2025-01-30
Requests page fails due to trying to modify a frozen string.

### LINKED ISSUE #611: MiniProfiler storage failure: can't modify frozen
MiniProfiler storage failure: can't modify frozen String: "<!DOCTYPE html>\n<html>\n  <head>\n    <title>Rack::MiniProfiler Requests</title>\n  </head>\n  <body>\n  </body>\n</html>\n"

Rails 7

Rack::MiniProfiler.config.auto_inject = true
Rack::MiniProfiler.config.show_total_sql_count = true
Rack::MiniProfiler.config.enable_advanced_debugging_tools = true
Rack::MiniProfiler.config.enable_hotwire_turbo_drive_support = true
Rack::MiniProfiler.config.profile_parameter = true
These are my settings, 
 Upon hitting the /requests urrl i am getting this error in console.Also turbo streams seems to be not loading new data in speed badge, as a matter of fact my speed badge is dissappearing after every turbo stream request. If anyone have solution to these issues. Pls respond

DropsOfSerenity:
Frozen strings cannot have `#insert` called on them, this was causing an error on the requests page. This PR fixes that issue.

Fixes #611

### DIFF
--- a/spec/integration/middleware_spec.rb
+++ b/spec/integration/middleware_spec.rb
@@ -27,7 +27,7 @@ describe Rack::MiniProfiler do
       get '/rack-mini-profiler/requests', {}, 'HTTP_ACCEPT_ENCODING' => 'gzip, compress'
 
       expect(last_response.body).to include('<title>Rack::MiniProfiler Requests</title>')
-      expect(last_response.body).to match('<body>\n  </body>')
+      expect(last_response.body).to match('<body>\n  <script async nonce="" type="text/javascript" id="mini-profiler"')
     end
   end
 
--- a/lib/mini_profiler.rb
+++ b/lib/mini_profiler.rb
@@ -329,7 +329,7 @@ module Rack
             )
           end
         elsif path == '/rack-mini-profiler/requests'
-          status, headers, body = [200, { 'Content-Type' => 'text/html' }, [blank_page_html]]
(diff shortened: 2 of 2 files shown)


coalest:
It looks iike the `blank_page_html` is the only place we are creating a frozen string (the other strings have string interpolations so they aren't frozen despite the frozen string literal comment), so potentially we could just add a `dup` to that method instead of calling `dup` on every fragment?

DropsOfSerenity:
> It looks iike the `blank_page_html` is the only place we are creating a frozen string (the other strings have string interpolations so they aren't frozen despite the frozen string literal comment), so potentially we could just add a `dup` to that method instead of calling `dup` on every fragment?

Excellent, that sounds good to me, been a bit busy, but I should be able to make the edit soon.

spec/integration/middleware_spec.rb
```diff
@@ -27,7 +27,7 @@ def app
       get '/rack-mini-profiler/requests', {}, 'HTTP_ACCEPT_ENCODING' => 'gzip, compress'
 
       expect(last_response.body).to include('<title>Rack::MiniProfiler Requests</title>')
-      expect(last_response.body).to match('<body>\n  </body>')
+      expect(last_response.body).to match('<body>\n  <script async nonce="" type="text/javascript" id="mini-profiler"')
```
nateberkopec, line 30: This is interesting. Without the .dup, it looks like we were not inserting the rmp script?

PR #631, opened 2025-01-31
Fix missing badge for Turbo Drive page loads

### LINKED ISSUE #612: Badge no longer displays after Turbo Drive page load w/ Turbo 8 (even w/ enable_hotwire_turbo_drive_support = true)
**Issue:** After upgrading to Turbo 8 the Rack Mini Profiler badge no longer displays after a Turbo Drive action with Rack::MiniProfiler.config.enable_hotwire_turbo_drive_support = true. If I revert back to Turbo 7 / Turbo Rails 1.5 the behavior works as expected.

**Expected Behavior:** When a page loads with Turbo Drive the Rack Mini Profiler badge displays.

**Actual Behavior:** When a page loads with Turbo Drive the Rack Mini Profiler badge does not display.

coalest:
Resolves #612 

Git bisect pointed to turbo-rails repo commit [d18b2ee](https://github.com/hotwired/turbo-rails/commit/d18b2eef50a17824005debd9adb1d9f6e823007f) being the cause of this bug. Previously Turbo Drive requests were displayed via the callback added to the `window.fetch`. That turbo-rails commit however change it so that a reference to `window.fetch` was stored in a variable when Turbo JS is loaded (before rack-mini-profiler has added the then callback) meaning that the callback to display those profiling request results doesn't get added to Turbo requests.

Adding a callback to `window.Turbo.fetch` is difficult as `window.Turbo` is a frozen object (unless we refactored to grab a reference before Turbo does by injecting some JS first), so instead a turbo event listener is used.


coalest:
Nevermind this only works for previous versions of Turbo, not the most recent.

webzorg:
I just tried this and it works for me? What did not work for you?

```
Rails 8.0.1
turbo-rails (2.0.11) (this one is latest)
```

coalest:
Whoops, when I tested this my browser was caching an old `includes.js` file without these changes.

I tried again and it does seem to work. Thanks for trying this out @webzorg

webzorg:
@coalest hehe least I can do, primarily, thank you for keeping this gem healthy and clean =D can't wait to change this to master =D `gem "rack-mini-profiler", path: "../rack-mini-profiler"`

pinzonjulian:
Hey! I'm giving it a try in the reproduction app I linked in the issue and it works but there's still a missing piece regarding new defaults in Turbo 8.

https://github.com/user-attachments/assets/dae5f49e-97fe-4765-be3f-1a9399f23a87

The issue here is due to the [default setting in Turbo Rails for prefetching links on hover](https://turbo.hotwired.dev/handbook/drive#prefetching-links-on-hover):

You'll notice in this video that:
- It still works for non-turbo requests
- On hover, mini-profiler does its job and changes the widget
- when you hover **and then click the link**, the badge disappears.

I'm also trying it out at my company and it doesn't seem to work so I'm trying to figure out what might be causing it. I'll report back with any findings

coalest:
@pinzonjulian I was also surprised by that behavior at first. It's an artifact due to how Turbo works.

Caching is disabled by RMP, which is why Turbo keeps fetching each of the partials on hover (even multiple times). So you might expect that even if you click on a link that has been prefetched, that Turbo would make another request once you perform the click (and you would see the badge after clicking). But Turbo caches the last request (and only the last request) and uses an event listener to add that cached response ([source](https://github.com/hotwired/turbo-rails/blob/main/app/assets/javascripts/turbo.js#L3287)). So if the link clicked matches the url of the last request, it just uses that cache ([source](https://github.com/hotwired/turbo-rails/blob/main/app/assets/javascripts/turbo.js#L3311)).

So the badge gets an additonal line item whenever a request gets sent, and when you click to a turbo link the badge stays if that link was clicked quickly (without hovered beforehand), otherwise that request gets added to the badge but clicking resets the badge to being empty again.

Regardless though of why it has this behavior, is this behavior ok or should it be fixed? If we can agree on what the desired behavior is, I can change it. Should the entire badge stay (with all the previous requests) when clicking. Should just the very last fetch request remain (the one for the link being clicked)?

coalest:
If it's a matter of keeping the badge with all requests, all we would have to do is remove the call to `pageTransition` when clicking a turbo link:

```diff
diff --git a/lib/html/includes.js b/lib/html/includes.js
index 1c7438d..c12fa0e 100644
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -548,7 +548,6 @@ var _MiniProfiler = (function() {
   var onTurboBeforeVisit = function onTurboBeforeVisit(e) {
     if(!e.defaultPrevented) {
       window.MiniProfilerContainer = document.querySelector('body > .profiler-results')
-      window.MiniProfiler.pageTransition()
     }
   }
```

Just keeping the prefetch would be a little more involved.

pinzonjulian:
From a UX standpoint, when I use RMP I expect the badge to never disappear. In the video I sent I carefully hover and don't click to show the issue but in real world usage, I would just click without even realising prefetching is happening and expect the badge to be there. 

It might get a tad confusing with prefetch requests piling up in the badge while you hover around the page but I think that if you're working on performance of a Turbo app, you know that's the case so I don't think it'd be an issue.

nateberkopec:
I'd rather have prefetch results in the speed badge and then have a way to ignore them, later (which may already be possible with current config settings).

nateberkopec:
Sorry, I realize my last comment wasn't helpful because that wasn't what you asked.

>Should the entire badge stay (with all the previous requests) when clicking? Should just the very last fetch request remain (the one for the link being clicked)?

The latter. Presumably they're the same request (right?) 

>It might get a tad confusing with prefetch requests piling up in the badge while you hover around the page

To be honest, I like the behavior in the video.

Do we have precedent around how the badge works re: Turbo requests in the past? I like the way the video works but I also don't want to break anybody's expectations as to what we did before.

leonvogt:
I'm really looking forward to this PR! 
It would be awesome to have better Turbo integration (again).

I feel the same as @pinzonjulian.
I would expect the badge to never disappear. I use prefetch in some of my apps and think it would be kind of confusing why the badge disappears after clicking on a link.

coalest:
I just pushed commit 6f7f37d which should preserve the last prefetched result in the badge. One drawback is that the workaround is pretty tightly coupled with Turbo (I use `Turbo.session.view.lastRenderedLocation.href` to check if the Turbo is going to send a request or use it's cache). Although we are already relying on Turbo-specific event listeners, so maybe that's not a big deal.

pinzonjulian:
I can confirm that it works well in the reproduction app!

https://github.com/user-attachments/assets/8ea1f60d-0c10-41fa-8b63-8c67aacdff96

Sadly I still haven't been able to make it work at my company; the badge just keeps disappearing. I'm guessing it's something on our side now that this is solved.

Thanks for tackling it!

coalest:
@pinzonjulian Bummer it's still not appearing on your company's app. I will take a look at turbo frames and turbo streams next to see if anything can be improved there.

emwalker:
This PR fixes the issue for me.

coalest:
@nateberkopec Would it be possible to get a code review on this PR? Or would you prefer a different maintainer to review it?

exterm:
Coming back two months later to similarly ask if anything is missing - I need this so would be happy to contribute - @nateberkopec ?

a-chris:
hello guys, any news on this? It would be awesome to have this working in the gem 🙏

thomaswitt:
I would also highly appreciate a merge and a release, as I am affected by the current behaviour.

leonvogt:
For about three months, I've been using the branch from @coalest without any problem.

```rb
gem "rack-mini-profiler", require: false, github: "coalest/rack-mini-profiler", branch: "fix-missing-badge-for-turbo-drive"
```

jclusso:
Any plan to merge this?

kbrock:
Is there a way to get this running on ruby 3.1?
I'm guessing that is the cause for the holdup? (not a maintainer, just guessing)

larouxn:
> Is there a way to get this running on ruby 3.1? I'm guessing that is the cause for the holdup? (not a maintainer, just guessing)

I don't think we even need to. Ruby 3.1 is EOL ([source](https://www.ruby-lang.org/en/downloads/branches/)) and has already been dropped from CI as of https://github.com/MiniProfiler/rack-mini-profiler/pull/653 in late July. I think this branch may simply need a rebase on `master` to get Ruby 3.1 CI to stop appearing/failing.

Additionally, I noticed that while this gem doesn't _test_ Ruby 3.1 anymore it still technically _supports_ it. I've opened a PR to officially drop Ruby 3.1 support, similar to what https://github.com/MiniProfiler/rack-mini-profiler/pull/632 did back in February, as Ruby 3.1 is EOL.
- https://github.com/MiniProfiler/rack-mini-profiler/pull/658

coalest:
This PR was strictly changing Javascript files (before I just fixed an unrelated linting error), so I highly doubt that this wouldn't have run on Ruby 3.1. Likely just a flaky CI run.

I'm assuming the reluctance to merge this PR is because the change is purely JS, and RMP doesn't have any testing infrastructure for testing front end behavior of the badge. :shrug:

larouxn:
> > Is there a way to get this running on ruby 3.1? I'm guessing that is the cause for the holdup? (not a maintainer, just guessing)
> 
> I don't think we even need to. Ruby 3.1 is EOL ([source](https://www.ruby-lang.org/en/downloads/branches/)) and has already been dropped from CI as of #653 in late July. I think this branch may simply need a rebase on `master` to get Ruby 3.1 CI to stop appearing/failing.
> 
> Additionally, I noticed that while this gem doesn't _test_ Ruby 3.1 anymore it still technically _supports_ it. I've opened a PR to officially drop Ruby 3.1 support, similar to what #632 did back in February, as Ruby 3.1 is EOL.
> 
> - https://github.com/MiniProfiler/rack-mini-profiler/pull/658

Dropping EOL Ruby 3.1 support has been merged, by the way.

thomaswitt:
Just a short ping - any merging progress here?

SamSaffron:
merging this, sorry it has taken so long

PR #632, opened 2025-02-03
Drop support for Ruby 2.7 and 3.0

tgxworld:
These versions are already EOL. Adds Ruby 3.4 to CI matrix.

### DIFF
--- a/rack-mini-profiler.gemspec
+++ b/rack-mini-profiler.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
     "CHANGELOG.md"
   ]
   s.add_runtime_dependency 'rack', '>= 1.2.0'
-  s.required_ruby_version = '>= 2.7.0'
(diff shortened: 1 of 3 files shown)


rack-mini-profiler.gemspec
```diff
@@ -41,7 +41,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop-discourse'
   s.add_development_dependency 'listen'
   s.add_development_dependency 'webpacker'
-  s.add_development_dependency 'rails', '~> 6.0'
+  s.add_development_dependency 'rails', '>= 7.1'
```
tgxworld, line 44: Need to bump `rails` dependency to resolve the following error when running rspec tests:

```
NameError:
  uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger
```

https://github.com/rails/rails/issues/54263

PR #633, opened 2025-02-03
Fix failing CI

tgxworld:
This commit makes the tests work when rack 3 is installed and broken rubocop rules.

### DIFF
--- a/rack-mini-profiler.gemspec
+++ b/rack-mini-profiler.gemspec
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 lib = File.expand_path('../lib', __FILE__)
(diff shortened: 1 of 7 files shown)


rack-mini-profiler.gemspec
```diff
@@ -41,7 +41,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop-discourse'
   s.add_development_dependency 'listen'
   s.add_development_dependency 'webpacker'
-  s.add_development_dependency 'rails', '~> 6.0'
+  s.add_development_dependency 'rails', '>= 7.1'
```
tgxworld, line 44: This resolves the `uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger (NameError)` error

https://github.com/rails/rails/issues/54263

PR #634, opened 2025-02-03
Truncate long profiler name in profiler popup

tgxworld:
### Before

Screenshot 2025-02-03 at 11 57 03 AM

### After

Screenshot 2025-02-03 at 12 39 41 PM


SamSaffron:
look at the truncating on the left (it leaves left and right bounds) ... can you match the algorithm,  I would say we have a tiny bit more room though so you can adjust to suite.

tgxworld:
@SamSaffron I've updated the solution to be CSS based instead.

SamSaffron (approved):

SamSaffron:
I am OK with this but 

/some/long/...../some/other 

would be nicer if possible.

tgxworld:
> /some/long/...../some/other
> 
> would be nicer if possible.

We wouldn't be able to rely on a CSS solution in this case. This is the current heuristic which we use for the badge

```
      <span class="profiler-name">
        {{? it.name.length >= 30 }}
          {{= it.name.substring(0,15) + "..." + it.name.slice(-15) }}
        {{??}}
          {{= it.name}}
        {{?}}
      </span>
```

SamSaffron:
yeah this would require js to be dynamic. I guess this is way better than what we have now so we just go with it

PR #638, opened 2025-02-04
Display record count in UI

### LINKED ISSUE #596: Display row counts in the UI
Captured by #585, but needs display to user

coalest:
Resolves #596

### Screenshot
Updated to show nothing when no AR records were instantiated:
image


coalest:
Not sure whether to return 0 or nothing when a query doesn't instantiate any Active Record instances. Thoughts?

nateberkopec:
Definitely nothing, as people who don't use ActiveRecord will be annoyed at all the 0 spam

nateberkopec (changes requested):
Should display nothing in case of 0

coalest:
Fixed in [552105f](https://github.com/MiniProfiler/rack-mini-profiler/pull/638/commits/552105f9d33eafd12b33bfda3a4d8c1ae9d3bd39) to not show anything if no AR records were instantiated.

nateberkopec:
Thanks @coalest!

Issue #613, opened 2024-05-09
Assets served from Rails deauthorizing client-side profiler


Blayr:
When trying to validate my 
```
Rack::MiniProfiler.config.authorization_mode = :allow_authorized
```
implementation, all my `/mini-profiler-resources/results` requests were failing even though in my ApplicationController I was doing `Rack::MiniProfiler.authorize_request` on every request. After some investigating I found that when loading my assets, they responded with a Set-Cookie header that was deleting my __profilin cookie.

![Screenshot 2024-05-09 at 4 31 57 PM](https://github.com/MiniProfiler/rack-mini-profiler/assets/17113158/873941a8-8290-449e-8104-9a2f2ba46ad0)
![Screenshot 2024-05-09 at 4 30 35 PM](https://github.com/MiniProfiler/rack-mini-profiler/assets/17113158/f7ca880c-d26b-4cb7-9965-3eded306a8ec)

Even though my assets path was included in the `skip_paths` automatically and the `skip_it` variable was set to true
the header to delete the cookie would still be set in `handle_cookie`, deauthorizing the client when the server loaded the asset.

https://github.com/MiniProfiler/rack-mini-profiler/blob/5e42a57acab20125f910c0f29d82f19e7220ff31/lib/mini_profiler.rb#L168C1-L180C10

```
      skip_it = matches_action?('skip', env) || (
        @config.skip_paths &&
        @config.skip_paths.any? do |p|
          if p.instance_of?(String)
            path.start_with?(p)
          elsif p.instance_of?(Regexp)
            p.match?(path)
          end
        end
      )
      if skip_it
        return client_settings.handle_cookie(@app.call(env))
      end
```

https://github.com/MiniProfiler/rack-mini-profiler/blob/5e42a57acab20125f910c0f29d82f19e7220ff31/lib/mini_profiler/client_settings.rb#L42C1-L56C10
```
      def handle_cookie(result)
        status, headers, _body = result


        if (MiniProfiler.config.authorization_mode == :allow_authorized && !MiniProfiler.request_authorized?)
          # this is non-obvious, don't kill the profiling cookie on errors or short requests
          # this ensures that stuff that never reaches the rails stack does not kill profiling
          if status.to_i >= 200 && status.to_i < 300 && ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - @start) > 0.1)
            discard_cookie!(headers)
          end
        else
          write!(headers)
        end


        result
      end
```

[...]

PR #641, opened 2025-02-12
Fix: Prevent request deauthorization when loading assets

iberianpig:
### Description

This PR fixes the problem described in [Issue #613](https://github.com/MiniProfiler/rack-mini-profiler/issues/613), where requests to the `/assets/` directory caused the `__profilin` cookie to be deleted, which in turn deactivated the client-side profiler.

### Changes Made

- **Prevent Cookie Deletion**: 
  - Updated the `handle_cookie` method to stop cookie deletion for asset requests by using `preserve_cookie: true`. This keeps the profiler's state when loading assets.

- **Skip Asset Requests**: 
  - Added `Rack::MiniProfiler.config.skip_paths = ['/assets']` to ignore asset requests during profiling checks. This is important because requests to `/assets` can still happen even if `Rails.config.public_file_server.enabled` is `false`.
  - When an asset is not found, the `config.assets.unknown_asset_fallback` option allows fallback handling, which can lead to extra requests to `/assets`. These paths were not fully considered in the dynamic skip paths configuration.

- **Rails Version Handling**: 
  - Updated the condition to check which asset server configuration to use for different Rails versions. For Rails 5.1.0 and later, the code checks both `public_file_server.enabled` and `unknown_asset_fallback`. For Rails 5.0.0 and 4.2.0, it keeps the existing behavior.

With these changes, the unwanted cookie deletion that caused the error `ActionController::RoutingError (No route matches [POST] "/mini-profiler-resources/results")` no longer happens.

### DIFF
--- a/lib/mini_profiler.rb
+++ b/lib/mini_profiler.rb
@@ -176,7 +176,8 @@ module Rack
         end
       )
       if skip_it
-        return client_settings.handle_cookie(@app.call(env))
+        # preserve_cookie to ensure we don't deauthorize the request (delete the auth cookie) when loading assets
+        return client_settings.handle_cookie(@app.call(env), preserve_cookie: true)
       end
 
       skip_it = (@config.pre_authorize_cb && !@config.pre_authorize_cb.call(env))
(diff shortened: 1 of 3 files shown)


SamSaffron:
ok seems good to me, thanks

PR #642, opened 2025-03-11
Emphasize note about storage in production

janko:
I glanced over this paragraph when I was reading the docs, because the blockquote de-emphasized it visually. Use new GitHub highlights to make it stand out.

Before:
Screenshot 2025-03-11 at 13 59 26

After:
Screenshot 2025-03-11 at 13 58 48


nateberkopec:
Thanks! This is super cool, wasn't aware it was possible.

PR #643, opened 2025-03-16
Only remove the first iteration of backtrace_remove

vprigent:
Our work setup is as follows:
We use docker for our rails app
We store our app within "/portal" in docker.

We also have a few folders called ".../portal/..." within the code which were getting swallowed by gsub.


kbrock (approved):
- This pretty much assumes that the `backtrace_remove` is on the left hand side of the file. (think this is a valid assumption)
- Wonder if a regexp for `backtrace_remove` would meet your needs.
- This may be a breaking change for people.

Having said all that, I think this is a great addition.

SamSaffron:
yeah this feels to risky to me, it is a breaking change

PR #645, opened 2025-06-08
Disable advanced_debugging_tools by default in development

benjaminwood:
The `enable_advanced_debugging_tools` setting is potentially dangerous, especially in development environments that are exposed to external networks. For example, if you run a Rails server bound to a public interface (e.g. `-b 0.0.0.0`), anyone on the local network can access sensitive information such as environment variables or memory debugging data. This risk is even greater if the application is exposed via a tunnel for webhook testing or similar purposes.

Although the README states that `enable_advanced_debugging_tools` is disabled by default, the current Railtie enables it in development mode. This change corrects that inconsistency and prioritizes safety by requiring explicit opt-in for these tools.

### DIFF
--- a/lib/mini_profiler_rails/railtie.rb
+++ b/lib/mini_profiler_rails/railtie.rb
@@ -59,7 +59,6 @@ module Rack::MiniProfilerRails
 
     # Install the Middleware
     app.middleware.insert(0, Rack::MiniProfiler)
-    c.enable_advanced_debugging_tools = Rails.env.development?
 
     if ::Rack::MiniProfiler.patch_rails?
       # Attach to various Rails methods


nateberkopec:
Couple of thoughts

1. This probably isn't the most-used feature in RMP
2. Local/dev ENV variables at most companies contain at least 1 sensitive/prod key IME
3. It's not possible to reliably detect in RMP whether or not we're exposed to the wider internet

So I'm 60/40 leaning towards "always disable". Would be nice to slap this in 4.0, wdyt @tgxworld

benjaminwood:
This didn't make it in to the 4.0 release, but I still think it's the right default. Can we ship it in the next release?

natevick:
@nateberkopec I thought this was changed at 4.0, did I misunderstand?

SamSaffron:
I am merging lots of secrets in envs these days

PR #646, opened 2025-06-13
feat: add ESC key to collapse expanded profiler results

mdayaram:
When profiler results are expanded, pressing ESC will collapse them back to the summary view. Integrates with existing ESC key handling for popups.

### DIFF
--- a/lib/html/includes.js
+++ b/lib/html/includes.js
@@ -214,6 +214,25 @@ var _MiniProfiler = (function() {
     return tempElement.children[0];
   };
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
I am sorry but this is too risky, in this state ESC may be something the app wants to handle. 

I think we need a different keyboard shortcut here that is unlikely to clash

mdayaram:
ah, that's fair. I can't think of an equally memorable shortcut that would make sense in this context, so I'll just go ahead and close this PR.

PR #648, opened 2025-07-11
Add CSP nonce support for flamegraph rendering

xrav3nz:
The flamegraph page was missing CSP nonce attributes on its inline script tags, causing Content Security Policy violations when nonces were configured, **leading to a blank page** when trying to view flamegraphs.

This PR extracts the nonce retrieval logic into a shared method and applies it consistently to both the main profiler UI and flamegraph pages.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
thanks @xrav3nz , sorry it has taken so long

PR #651, opened 2025-07-24
CI: Configure setup-ruby to use a compatible RubyGems in order to build green on Ruby 3.1

olleolleolle:
This PR configures setup-ruby to use the latest-compatible RubyGems that for the current running Ruby.

This is an attempt to make mini_racer install correctly, on Ruby 3.1... which panned out, we are now building green 🟢 again.


nateberkopec:
Thanks for all the clean up @olleolleolle 🙇 You're a hero of Ruby OSS!

PR #657, opened 2025-09-16
Fix simple sql bind parameters throwing errors

kbrock:
```
FATAL -- : Error caught: [NoMethodError] undefined method `name' for an instance of String
rack-mini-profiler/lib/mini_profiler.rb:100:in `block in binds_to_params'
rack-mini-profiler/lib/mini_profiler.rb:100:in `map'
rack-mini-profiler/lib/mini_profiler.rb:100:in `binds_to_params'
rack-mini-profiler/lib/mini_profiler_rails/railtie.rb:114:in `block in initialize!'
rack-mini-profiler/lib/mini_profiler_rails/railtie.rb:146:in `block in subscribe'
```

Old Rails
=========

**It works.**

The `binds` parameters, passed to sql are of the following forms:

- `["parameter_name", "parameter_value"]`
- `ActiveRecord::Relation::QueryAttribute`

Rails 7.1 (or maybe 7.0?)
==============

**It fails**

The `binds` parameters, passed to sql are of the following forms:

- `["parameter_name", "parameter_value"]`
- `ActiveRecord::Relation::QueryAttribute`
- Simple class like `Date` or `String` <=== problem

The simple types caused an issue when trying to call `String#name` or `String#value`. See error above.

After this commit
=================

**It works**

It is able to handle when simple objects are passed into rails
If a password is passed as a simple parameter, that will get leaked,
since the name of the parameter is no longer passed by rails.

### DIFF
--- a/lib/mini_profiler.rb
+++ b/lib/mini_profiler.rb
@@ -96,8 +96,15 @@ module Rack
 
       def binds_to_params(binds)
         return if binds.nil? || config.max_sql_param_length == 0
-        # map ActiveRecord::Relation::QueryAttribute to [name, value]
-        params = binds.map { |c| c.kind_of?(Array) ? [c.first, c.last] : [c.name, c.value] }
+        params = binds.map do |c|
+          if c.kind_of?(Array)
+            [c.first, c.last]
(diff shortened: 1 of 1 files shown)


SamSaffron:
OK I have a red test, merging!

PR #658, opened 2025-09-16
Drop support for Ruby 3.1 as it is EOL

larouxn:
Ruby 3.1 is EOL as of ~earlier this year~ last year ([source](https://www.ruby-lang.org/en/downloads/branches/)) and as such we should probably drop support like https://github.com/MiniProfiler/rack-mini-profiler/pull/632 did for Ruby 3.0. Additionally this repo/gem's CI doesn't even test Ruby 3.1 anymore as of https://github.com/MiniProfiler/rack-mini-profiler/pull/653 from July 31st of this year. Thus, it's in fact a bit risky to still allow Ruby 3.1.

Inspired by https://github.com/MiniProfiler/rack-mini-profiler/pull/631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # CHANGELOG
 
+- [BREAKING CHANGE] Ruby version 3.2.0 or later is required. [#632](https://github.com/MiniProfiler/rack-mini-profiler/pull/658)
+
 ## 4.0.1 - 2025-07-31
 
 - [FIX] Ensure Rack 2 / 3 cross compatibility [#653](https://github.com/MiniProfiler/rack-mini-profiler/pull/653)
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ If you feel like taking on any of this start an issue and update us on your prog
 
 ## Installation
 
-Install/add to Gemfile in Ruby 3.1+
+Install/add to Gemfile in Ruby 3.2+
 
 ```ruby
 gem 'rack-mini-profiler'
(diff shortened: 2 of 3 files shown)


kbrock:
The bookworm (oldstable) release of Debian uses Ruby 3.1 as its default and it is supported until 2026.

Not saying that we have to support Debian, but I've been using Debian as a gauge for the gems that I support. (I don't run Debian, it was just brought up to me on another gem and the request seemed sound to me)

larouxn:
That's an interesting a fair point regarding Debian. While I would prefer to drop support for Ruby 3.1 as it's EOL, I am not a maintainer nor (yet) contributor to this gem so I'll leave it up to such people to decide.

That said, if we do intend to continue allowing Ruby 3.1 installs until 2026 we should add it back to CI.

https://github.com/MiniProfiler/rack-mini-profiler/blob/92610caa0051510530da5def0602e5e97683f824/.github/workflows/ci.yml#L29

kbrock:
Good point.

To be clear, I do not maintain this gem. (have contributed for 9 years)
That was just something I threw out there based upon comments given to me when I tried to drop older versions of ruby

larouxn:
:wave: Any chance we can get this merged? Ruby 3.1 is now over 9 months passed EOL.

- https://www.ruby-lang.org/en/downloads/branches/
- https://endoflife.date/ruby

Heck at this rate we're coming up on Ruby 3.2 EOL at the end of March. :sweat_smile: 

CC @nateberkopec, @olleolleolle, @Flink 🙏

larouxn:
👋 Checking in here. Ruby 3.1 has been EOL for over 1 year now. I think we should, especially now, move forward with this change.

We could even require Ruby 3.3 at this point as 3.2 is now EOL as well, but that just happened a few weeks ago so maybe leave as is for now.

Flink (approved):
Sorry for the delay! You’re right, it’s probably a good idea 🙂

Flink:
Oh, I can approve but not merge :sweat_smile: :see_no_evil:

PR #660, opened 2026-03-19
CI: Tell dependabot to update GH Actions

olleolleolle:
This PR adds a configuration to tell Dependabot to update the versions used for some well-known GitHub Actions, such as `checkout`, in order to avoid deprecation warnings when those get too old. [Documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


kbrock:
Um, is this failing because rubocop is not nailed down.

To support multiple Rails versions, do we want to enforce `Rails.env.local?`? It was introduced in Rails 7.1.

nateberkopec:
We're also failing a bunch b/c of an old memcached version. I'm altering CI to update.

